PR #2, opened 2016-02-08
Update the code so it's fully compatible with Gravity PDF 4.0 RC2

blueliquiddesigns:
Notable changes include:

- Greatly simplifying the PDF generation command
- Tweak the checks used when using the GPDFAPI (is_wp_error is preferred)
- Updated version check to ensure commands aren't run on a version of Gravity PDF that doesn't support them
- Ensured v3 compatibility isn't run when using v4.

Note: Gravity PDF 4.0 RC2 has not yet been released to our testers. It's currently in development.

### DIFF
--- a/includes/steps/class-step-approval.php
+++ b/includes/steps/class-step-approval.php
@@ -279,30 +279,42 @@ class Gravity_Flow_Step_Approval extends Gravity_Flow_Step {
 
 
 		// Support for Gravity PDF 4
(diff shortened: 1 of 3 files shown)


blueliquiddesigns:
FYI, 4.0-RC2 was just released to our user-base for testing. 

More details can be found here: http://us2.campaign-archive2.com/?u=552d40c63f6f2b4503e3d8d77&id=9a83f6822c

PR #11, opened 2017-02-18
add support for repeat reminders

stevehenty:
Adds a new setting type (checkbox_and_container) which hides the contents of the container depending on the value of the checkbox.

Updates the assignee reminder setting to use the new setting type and adds settings for repeat reminders.

Updates the cron task to send repeat reminders.


stevehenty:
Refactored settings_checkbox_and_text() and settings_checkbox_and_textarea() to use settings_checkbox_and_container()

PR #13, opened 2017-02-23
prevent dynamic choices from being added to the form meta

stevehenty:
Currently the choices for the User and Role fields are getting saved into the Form meta in the Form Editor. This causes the conditional logic and field filters to use a dropdown with a list of choices that gets out of date. This fix removes the choices when the field settings are opened and prevents them from being added when the field is first created. The conditional logic and field filters will display a text box instead of a dropdown.


stevehenty:
- Added fix during upgrade
- Fixed choices affected in other fields

PR #15, opened 2017-02-26
Fix two issues with post images

richardW8k:
- fix post image merge tags for existing images
- remove the fields old post image when a new image is saved to the
entry


stevehenty:
I'm seeing an issue with the post content not updating. The form has 3 post image fields, one User Input step with one editable post image field - the other two are displayed. The post body isn't adding the new image although it's is attached to the post correctly.

richardW8k:
Is the post body field editable? The post content is only being updated if a new value is supplied for the post body field. We could have it also update if the field isn't editable and has a content template containing merge tags.

stevehenty:
It's editable but I left the value unchanged. So, yeah, I agree it would make sense to update the post content if it contains merge tags.

richardW8k:
Updated to always process the content template. If a template isn't configured then the post content will only be updated if the body field has a value.

PR #22, opened 2017-03-19
User input email tabs

richardW8k:
- Added tabs for configuring emails to be sent when the entry is updated and the step remains in progress and when the step is completed.
- Updated how notification tabs are handled in form-settings.js to reduce duplication.


stevehenty:
It's working well. Just a couple of observations:
- The merge tags icon is not available on the editors.
- The discussion field merge tag is producing some encoded output:
http://d.pr/i/T4Ou/2hCQdzxD

stevehenty:
Merge tag issue fixed in GF 2.2

richardW8k:
Renamed the User Input step 'Default Status Option' setting to 'Save Progress Option' and changed the setting type from radio buttons to drop down. The In Progress notification tab is disabled if the disabled choice is selected.

PR #24, opened 2017-04-01
Instructions unfiltered html

stevehenty:
- Added validation for the checkbox and textbox setting.
- Added the gravityflow_sanitize_instructions filter.
- Removed wp_kses_post() from the instructions output.

### DIFF
--- a/change_log.txt
+++ b/change_log.txt
@@ -1,4 +1,4 @@
-
+- Added support for unfiltered HTML in the instructions setting for users with the unfiltered_html capability.
 - Added the ability to configure emails to be sent when a User Input step updates the entry and remains in progress and when the step is completed.
 - Fixed an issue with the notification message when the Discussion field merge tag content exceeded the max line length.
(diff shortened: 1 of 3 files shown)


richardW8k:
I'm not seeing the unfiltered_html capability check mentioned in the change log, was that lost in a merge?

stevehenty:
It seems to be there:
https://github.com/gravityflow/gravityflow/blob/instructions-unfiltered-html/change_log.txt

richardW8k:
Right, but the `current_user_can( 'unfiltered_html' )` check is missing from maybe_sanitize_instructions().

stevehenty:
Ah, I misunderstood, I thought you meant it was missing from the changelog.

The instructions now allow unfiltered HTML to be displayed - so no check for the current user's capabilities is required. To keep it secure, the instructions setting is now validated and won't allow unfiltered HTML unless the admin can unfiltered_html.

PR #29, opened 2017-05-07
adding field context properties

stevehenty:
Adds editable and display content properties so third party add-ons that don't use GF_Field can provide support for the User Input step more easily.


richardW8k:
Should $field->gravityflow_is_display_field also be set in Gravity_Flow_Entry_Detail::fields()?

stevehenty:
Yes, thanks, I was only thinking about the User Input step but it'll be useful for the workflow detail page too.

PR #35, opened 2017-06-12
Fixed logic scripts js error with chained select field

richardW8k:
Fixes a "gf_form_conditional_logic is not defined" JS error which occurred on the detail page with the user input step when the chained select field is editable and conditional logic is not enabled for the step.


stevehenty:
I'm still seeing the gf_form_conditional_logic is not defined error in both the shortcode and admin UI.

Also, it looks like the Chained Selects scripts aren't loading in the admin UI. I'm seeing the following error in the console: GFChainedSelects is not defined.

richardW8k:
Weird, I'm not seeing the gf_form_conditional_logic error with the latest versions of this branch and chained selects and my test form. Could you send me the form you are testing with?

It looks like the GFChainedSelects error was introduced this week when the [enqueue condition was changed](https://github.com/gravityforms/gravityformschainedselects/commit/8beda1415b90b188120c5e0c0cfecbc38cb4a1b9).

I'll have word with Dave about it. It should probably use the frameworks is_{page}() helpers rather than is_admin().

richardW8k:
The field having logic configured was still causing conditional_logic.js to be included, however, logic being disabled on the step had prevented the logic init scripts from being included. 

Chained Selects is using the gform_input_change hook which is also used in conditional_logic.js where it uses the gf_form_conditional_logic variable from the logic init scripts.

The latest commit ensures that the gf_form_conditional_logic variable is defined when conditional logic is disabled for the step, conditional_logic.js has already been included, and the form has an editable Chained Select field.

Longer term, and most likely as part of the entry editor/editable fields changes, we should look into filtering the form as early as Gravity_Flow::enqueue_form_scripts() so the form passed to GFFormDisplay::enqueue_form_scripts() does not contain irrelevant fields and conditional logic is removed etc.

stevehenty:
The shortcode is working well with field conditional logic enabled. With field conditional logic disabled I'm seeing a gf_form_conditional_logic is not defined error when changing selects.

In the admin UI I'm seeing a gformChainedSelectData is not defined error when changing selects with field conditional logic both enabled and disabled.

PR #36, opened 2017-06-13
Added discussion field modifiers

richardW8k:
Added the first, last and limit modifiers.


stevehenty:
Looking good. Just thinking that "latest" might be more precise than "last" but it's not a big deal if you want to go with that.

While testing I noticed that the discussion field merge tag isn't rendering anything in the instructions. Can we fix that here?

PR #38, opened 2017-06-19
Gforms 2.3 compatibility

stevehenty:
add support for Gravity Forms 2.3


includes/models/class-activity.php
```diff
@@ -22,17 +22,21 @@ public static function get_lead_table_name() {
 		return GFFormsModel::get_lead_table_name();
 	}
 
+	public static function get_entry_table_name() {
```
richardW8k: Rather than having this here, could Gravity_Flow_Activity::get_events() use Gravity_Flow_Common::get_entry_table_name() instead?

PR #42, opened 2017-07-07
Fix redirect bug when incorrect HTTP_HOST is passed / set (this is user supplied)

jakejackson1:
Remove reliance on `$_SERVER` variables to determine protocol and site URL and use WordPress' in-built `home_url()` method instead (`home_url()` will point to the front end WordPress site, rather than where the WordPress files are located - `site_url()`)

### DIFF
--- a/class-gravity-flow.php
+++ b/class-gravity-flow.php
@@ -4773,14 +4773,13 @@ AND m.meta_value='queued'";
 
 				if ( ! empty( $token ) && ! isset( $token['scopes']['action'] )&& ! is_user_logged_in() ) {
 					// Remove the token from the URL to avoid accidental sharing.
-					$secure = ( 'https' === parse_url( site_url(), PHP_URL_SCHEME ) );
+					$secure = ( 'https' === parse_url( home_url(), PHP_URL_SCHEME ) );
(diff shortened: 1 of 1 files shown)


stevehenty:
Thanks for this, Jake. There is certainly an issue we need to address here. The current code will not work as expected in some situations for example on a server behind a reverse proxy, or multisite on nginx with domain mapping. The trouble with home_url() is that it can point to a completely different installation. So, I think we're going to have to solve this with a filter like Gravity Forms does in these cases.

jakejackson1:
Hi Steve,

In what circumstances would the `home_url()` point to a completely different site than the one currently being accessed? It was my understanding that it only differs from the `site_url()` when you move the [WordPress files to its own directory](https://codex.wordpress.org/Giving_WordPress_Its_Own_Directory#Method_II_.28With_URL_change.29).

stevehenty:
I've been thinking about this a lot and I think there's a case for adding a better way to return the root URL along with some checks for exceptions, like rest_url().

To answer your question, some sites use it to link to the main site so it's not always reliable. But then, nothing is reliable. 

I'll close this for now but I'm happy to help the customer directly if they want to open a ticket.

stevehenty:
Re-opening and merging to get it into the next release. We can improve another time.

PR #51, opened 2017-07-22
Members 2.0+ Integration

richardW8k:
Updates the Members integration so that versions 2.0+ will group all the Gravity Flow and extension capabilities together on one tab. The core capabilities will display human readable labels. Extensions will continue to display the capability until get_members_capabilities() is overridden with the human readable labels.


richardW8k:
Also fixes two other issues:
 - $original_entry not being passed to the gform_after_update_entry hook on the user input step
 - Multi Select fields using the json storageType not being populated with the existing entry value on the user input step

PR #55, opened 2017-08-05
add raw body and headers settings

stevehenty:
Adds PATCH to the list of request methods in the request method setting of the Outgoing Webhook step settings.
Adds Headers to the Outgoing Webhook step settings.
Adds the Raw body setting to the OUtgoing webhook step settings.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -79,7 +79,7 @@ before_script:
 - grunt --version
 - git --version
 - svn --version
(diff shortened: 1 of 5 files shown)


richardW8k:
When selecting add custom key in the request headers setting I'm seeing a layout issue with the reset button: https://www.dropbox.com/s/gf85w0oe1qfzj2b/Screenshot%202017-08-06%2009.51.13.png?dl=0

PR #57, opened 2017-08-13
Add modifier to limit user notes to most recent occurrence of the specified step

richardW8k:
re: [HS#4305](https://secure.helpscout.net/conversation/412406037/4305/?folderId=516726)

When using the step_id modifier with the workflow_note merge tag all the user submitted notes with the specified step id will be returned. For the average workflow that is fine, however, if the workflow includes loops so a step can occur multiple times the notes from all occurrences are currently output.

The all_occurrences modifier when set to false will return only those notes belonging to the most recent occurrence of the specified step.

### DIFF
--- a/change_log.txt
+++ b/change_log.txt
@@ -3,7 +3,7 @@ Added the gravityflow_step_assignees filter.
 Added PATCH to the list of request methods in the request method setting of the Outgoing Webhook step settings.
 Added Headers to the Outgoing Webhook step settings.
 Added the Raw body setting to the OUtgoing webhook step settings.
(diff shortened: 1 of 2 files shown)


stevehenty:
In the case of loops, it seems like only the latest note is relevant and that the previous notes would not be needed so perhaps the default should be false instead of true? Also, I'm not sure about the name of the modifier. How about "history" or instead of "all_occurences"? Or perhaps "limit" with a number like the Discussion Field?

stevehenty:
The refactoring of the loop has knocked it out so the history attribute has no effect now.

PR #63, opened 2017-09-17
Oauth1 Addition

LiamBailey:
Note - latest changes from main fork pulled in this morning and my work rebased on top. I then completely overwrote the plugin on the test server with the plugin as submitted in this pull request and tested  - all oauth functionality is working as expected

What doesn't work with the WP API - is: as discussed the raw body with no json according to the response this is processed as a GET request to the endpoint and sends back the object being requested in the response.

Also - the All Fields setting doesn't work - according to the response this is because a "status" parameter is being sent with the value "active" most likely Gravity Forms form meta data, and WP API is looking for this to be post_status with accepted params being "published", "draft" etc.


stevehenty:
Development continued in the pr/63 branch.

PR #64, opened 2017-09-26
Move editable fields processing and related functionality out of the User Input step

richardW8k:
This moves the functionality for validating editable fields, updating the entry, and updating the post which was created during form submission out of the User Input step and into the new Gravity_Flow_Entry_Updater class.

This change makes it easier to implement editable fields on other step types in the future.


richardW8k:
[Move Entry Edit code out of the User Input step into the entry editor.](https://trello.com/c/dQc1n7dq/260-move-entry-edit-code-out-of-the-user-input-step-into-the-entry-editor)

PR #65, opened 2017-10-04
Fix User Step default_value for field conditional logic

Idealien:
This ensures new user input steps for forms with conditional logic fields or pages have the 'Enable field conditional logic' checkbox active by default. Does not impact existing user input steps that had already stored value on / off.


richardW8k:
[Activate the "enable field conditional logic" setting on the User Input step by default for new steps.](https://trello.com/c/DRABEMOz/309-activate-the-enable-field-conditional-logic-setting-on-the-user-input-step-by-default-for-new-steps)

PR #66, opened 2017-10-06
Enabling Form Title column for Inbox with multiple forms set via the gravityflow_inbox_args filter

Idealien:
The filter gravityflow_inbox_args provides method to limit the shortcode to an array of forms. But the check whether to display Form Name as column in the inbox assumes the form_id argument is empty. This PR expands it to also check/allow array values to trigger Form Name column display.

https://gist.github.com/Idealien/aaeb8538adcf28d16f721ca4c44f249b is example of filter use that would not  display Form Name column without the PR.

### DIFF
--- a/includes/pages/class-inbox.php
+++ b/includes/pages/class-inbox.php
@@ -201,7 +201,7 @@ class Gravity_Flow_Inbox {
 			$columns['actions'] = '';
 		}
 
-		if ( empty( $args['form_id'] ) ) {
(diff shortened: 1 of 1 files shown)


stevehenty:
The shortcode doesn't support multiple forms because the field IDs will be different on each form and I'm concerned about the support requests. Adding support via the filter is fine though.

stevehenty:
I've changed the title of the PR to reflect my last comment.

PR #71, opened 2017-10-30
Add entry_id attribute for shortcode for inbox / status to direct link

Idealien:
Attributes in [The Shortcode docs](http://docs.gravityflow.io/article/36-the-shortcode) for Inbox / Status will need to be updated to include entry_id. Noting that any page where lid is in querystring params takes precedence over the shortcode attribute.

### DIFF
--- a/class-gravity-flow.php
+++ b/class-gravity-flow.php
@@ -3425,9 +3425,15 @@ PRIMARY KEY  (id)
 
 			$args = array_merge( $defaults, $args );
 
-			if ( rgget( 'view' ) == 'entry' ) {
+			if ( rgget( 'view' ) == 'entry' || ! empty( $args['entry_id'] ) ) {
 
 				$entry_id = absint( rgget( 'lid' ) );
+
+				if ( empty( $entry_id ) ) {
+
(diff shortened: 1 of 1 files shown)


class-gravity-flow.php
```diff
@@ -4055,7 +4070,7 @@ public function shortcode( $atts, $content = null ) {
 					wp_enqueue_script( 'gravityflow_entry_detail' );
 					wp_enqueue_script( 'gravityflow_status_list' );
 
-					if ( rgget( 'view' ) ) {
+					if ( rgget( 'view' ) || ! empty( $a['entry_id'] ) ) {
```
stevehenty: Should this be $entry_id?

class-gravity-flow.php
```diff
@@ -4028,6 +4034,15 @@ public function shortcode( $atts, $content = null ) {
 
 			$entry_id = absint( rgget( 'lid' ) );
 
+			if ( empty( $entry_id ) ) {
+
+				if ( ! empty( $a['entry_id'] ) ) {
```
stevehenty: This conditional can be combined with its parent.

PR #73, opened 2017-11-08
Save Progress - Add Submit Button option + logic

Idealien:
This extends the user input step type Save Progress option drop-down with an additional 'Two Button' type. It displays two form submit buttons for Save Progress and Submit instead of the radio + form submit button.

Two areas where review/feedback is useful:

1. **Filters in display_update_button** - Whether should match the filter types (field label and entire button) of the other scenarios and/or apply same filters - like gravityflow_update_button_text_user_input - which have "identical" purpose.

1. **maybe_enable_update_button** - Whether this approach of enabling non-existent buttons is ok/preferred vs checking for save progress option and separate script tag blocks.

### DIFF
--- a/includes/steps/class-step-user-input.php
+++ b/includes/steps/class-step-user-input.php
@@ -148,13 +148,14 @@ class Gravity_Flow_Step_User_Input extends Gravity_Flow_Step {
 			array(
 				'name'          => 'default_status',
 				'type'          => 'select',
-				'label'         => __( 'Save Progress Option', 'gravityflow' ),
+				'label'         => __( 'Save Progress', 'gravityflow' ),
 				'tooltip'       => __( 'This setting allows the assignee to save the field values without submitting the form as complete. Select Disabled to hide the "in progress" option or select the default value for the radio buttons.', 'gravityflow' ),
 				'default_value' => 'hidden',
 				'choices'       => array(
 					array( 'label' => __( 'Disabled', 'gravityflow' ), 'value' => 'hidden' ),
 					array( 'label' => __( 'Radio buttons (default: In progress)', 'gravityflow' ), 'value' => 'in_progress' ),
 					array( 'label' => __( 'Radio buttons (default: Complete)', 'gravityflow' ), 'value' => 'complete' ),
+					array( 'label' => __( 'Two buttons ("Save Progress" and "Submit")', 'gravityflow' ), 'value' => 'two_buttons' ),
 				),
 			),
 			array(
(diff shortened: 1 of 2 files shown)


stevehenty:
Let's put a space between the two buttons and change the "Save Progress" button to "Save".

stevehenty:
1. it looks like they match which makes sense.
2. the objective is to make it as easy as possible to extend the User Input class. do you think it's easy like this or do you think it needs to be separated?

Idealien:
Re: #2 - It's now checking value of default_status in PHP to determine which JS to inline. While equally "easy" to extend in previous form, I think this reduces potential edge case issues and only has JS trying to fire on elements which exist on the page.

includes/steps/class-step-user-input.php
```diff
@@ -867,6 +879,7 @@ public function display_update_button( $form ) {
 				* @params string $save_progress_button
 				*/
 				echo apply_filters( 'gravityflow_save_progress_button_user_input', $save_progress_button );
+				echo '&nbsp;&nbsp;';
```
stevehenty: Let's use CSS for this.

includes/steps/class-step-user-input.php
```diff
@@ -775,17 +775,29 @@ public function maybe_enable_update_button( $can_update ) {
...
+			?>
+			<script>
+				(function (GFFlowInput, $) {
+					$(document).ready(function () {
+						$('#gravityflow_update_button').prop('disabled', false);
+					});
+				}(window.GFFlowInput = window.GFFlowInput || {}, jQuery));
+			</script>
+			<?php
+		}
```
stevehenty: You can reduce some duplication by putting the PHP conditional inside the JS function.

GFFlowInput looks like a typo. Could you change that to GravityFlowUserInput?

includes/steps/class-step-user-input.php
```diff
@@ -871,15 +863,14 @@ public function display_update_button( $form ) {
 				$save_progress_button_text   = apply_filters( 'gravityflow_save_progress_button_text_user_input', $save_progress_button_text, $form, $this );
 				$save_progress_button_click  = "jQuery('#action').val('update'); jQuery('#gravityflow_status_hidden').val('in_progress'); jQuery('#gform_{$form_id}').submit(); return false;";
 				$save_progress_button_id     = 'gravityflow_save_progress_button';
-				$save_progress_button        = '<input id="' . $save_progress_button_id . '" disabled="disabled" class="button button-large button-primary" type="submit" tabindex="4" value="' . $save_progress_button_text . '" name="in_progress" onclick="' . $save_progress_button_click . '"/>';
+				$save_progress_button        = '<input id="' . $save_progress_button_id . '" disabled="disabled" class="button button-large button-secondary" type="submit" tabindex="4" value="' . $save_progress_button_text . '" name="in_progress" onclick="' . $save_progress_button_click . '" style="margin-right:10px;" />';
```
stevehenty: We need to avoid inline CSS. This can be moved to entry-detail.css and front-end.css.

PR #76, opened 2017-11-18
Add invoice sent status

stevehenty:
Adds the sent status of a sliced invoice.


includes/steps/class-step-feed-slicedinvoices.php
```diff
@@ -266,7 +266,13 @@ public function workflow_detail_box( $form, $args ) {
 				}
 
 				if ( class_exists( 'Sliced_Shared' ) ) {
-					echo sprintf( '%s: %s<br><br>', esc_html__( 'Status', 'gravityflow' ), esc_html( Sliced_Shared::get_status( $invoice->ID, 'invoice' ) ) );
+					$sent_status = '';
+
+					if ( class_exists( 'Sliced_Shared' ) && Sliced_Invoice::get_email_sent_date( $invoice->ID ) ) {
```
richardW8k: Should this class_exists check be for Sliced_Invoice instead of Sliced_Shared?

PR #78, opened 2017-11-28
Step Highlight

Idealien:
Ready for an initial code review for the Step Highlight - Color Setting.

- All step types get new setting 'Highlight - Highlight this step' that is not selected by default.
- Part 1 PR has step_highlight_type hidden / defaulted to color
- New JS dependency to wp-color-picker on step edit screen
- New filter - gravityflow_step_highlight_color_inbox - that could be used to change from a step-defined color to a field-derived status/priority. Passes $form['id'], $entry as additional parameters to the step_highlight_color to return (color hexcode)

Defined step highlight color can display in 3 locations:
1. Inbox Shortcode - new column step_highlight - off by default as true/false
1. Workflow inbox - on by default args, can be overridden through gravityflow_inbox_args
1. Workflow Step List

PR also includes small inactive plumbing for part 2 of text, icon and upload svg step highights which a future PR will complete and activate.

### DIFF
--- a/class-gravity-flow.php
+++ b/class-gravity-flow.php
@@ -1828,6 +1828,8 @@ PRIMARY KEY  (id)
 		/**
 		* Prepare the step_highlight composite settings to be accessible for every field in the composite
 		*
+		* @since 1.9.2
+		*
 		* @param array $field The step_highlight field
 		*
 		* @return array
@@ -1872,6 +1874,8 @@ PRIMARY KEY  (id)
 		* Generate the step_highlight composite setting container
 		*
 		* The container will be displayed or hidden depending on the value of the step_highlight checkbox field.
+		*
+		* @since 1.9.2
 		*
 		* @param array $field The step_highlight field
 		*/
--- a/class-gravity-flow.php
+++ b/class-gravity-flow.php
@@ -1810,23 +1810,39 @@ PRIMARY KEY  (id)
 			<?php
 		}
 
-		public function settings_step_highlight( $field, $echo = true ) {
+		/**
+		 * Prepare and render markup for the step_highlight composite setting.
+		 *
+		 * The container will be displayed or hidden depending on the value of the step_highlight checkbox field.
+		 *
+		 * @since 1.9.2
+		 *
+		 * @param array $field
+		 */
+		public function settings_step_highlight( $field ) {
 			$field = $this->prepare_settings_step_highlight( $field );
 
-			return $this->settings_step_highlight_container( $field, $echo );
+			return $this->settings_step_highlight_container( $field );
 		}
 
+		/**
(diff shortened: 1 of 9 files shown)


includes/pages/class-inbox.php
```diff
@@ -262,7 +270,39 @@ public static function display_entry_row( $args, $entry, $columns ) {
...
+		/**
+		 * Allow the Step Highlight colour to be overridden.
+		 *
+		 * @since 1.9.2
+		 *
+		 * @param string $highlight The highlight color (hex value) of the row currently being processed.
+		 * @param int $form['id'] The ID of form currently being processed.
+		 * @param array $entry The entry object for the row currently being processed.
+		 */
+		$step_highlight = apply_filters( 'gravityflow_step_highlight_color_inbox', $step_highlight_color, $form['id'], $entry );
```
stevehenty: `$step_highlight`is not used later

css/frontend.css
```diff
@@ -334,7 +334,10 @@ table.entry-products col.entry-products-col4{
         width:100%!important;
         border-top:1px solid #ddd !important;
     }
-
+    
+    table#gravityflow-inbox td[data-label="Highlight"] {
+        display: none !important;
+    }
```
stevehenty: Is this block still necessary?

class-gravity-flow.php
```diff
@@ -1801,6 +1810,111 @@ public function settings_expiration( $field ) {
...
+			$field['settings']['step_highlight_color'] = $step_highlight_color;
+
+			return $field;
+		}
+
+		/**
+		* Generate the step_highlight composite setting container
+		*
+		* The container will be displayed or hidden depending on the value of the step_highlight checkbox field.
+		*
```
stevehenty: Missing since tag

class-gravity-flow.php
```diff
@@ -1801,6 +1810,111 @@ public function settings_expiration( $field ) {
...
+		 */
+		public function settings_step_highlight( $field ) {
+			$field = $this->prepare_settings_step_highlight( $field );
+
+			return $this->settings_step_highlight_container( $field );
+		}
+
+		/**
+		* Prepare the step_highlight composite settings to be accessible for every field in the composite
+		*
```
stevehenty: missing since tag

Idealien: Fixed.

class-gravity-flow.php
Idealien: Fixed.

css/frontend.css
Idealien: No. Removed.

includes/pages/class-inbox.php
Idealien: Corrected. Also added the @return in doc block.

PR #84, opened 2017-12-05
Escape new step URL

stevehenty:


richardW8k:
Should it be using esc_url instead of esc_html?

PR #89, opened 2017-12-12
Acceptance Test for Save Progress Types

Idealien:
Numbered as 0023 b/c PR #85 has 0022 in queue also


stevehenty:
The `waitFor*` functions are necessary but they don't assert anything. If you add functions like `see` the report will display more assertions.

PR #90, opened 2017-12-17
Add the workflow_fields merge tag

richardW8k:
Adds support for the {workflow_fields} merge tag which will output the steps display and/or editable fields using the same table based output as the all_fields merge tag.

The following modifiers can be set to true or false to customize the output:

- empty - should empty fields be included? default is false
- value - should choice based fields output the choice value instead of the choice text? default is false
- admin - should the fields admin label be used instead? default is false
- editable - should the editable fields be included? default is true
- display - should the display fields be included? default is true

{workflow_fields: display=false}


stevehenty:
Do `is_display_field()` and `is_editable_field()` need to be in different classes?

richardW8k:
Nope, at one point I did have them both in Gravity_Flow_Step. If I recall correctly one or more acceptance tests failed when the detail page was displayed after the workflow was complete so I moved is_display_field() to Gravity_Flow_Common to get the tests passing again and forgot to move the other.

PR #93, opened 2018-01-10
4933 translation filter

Idealien:
On the Entry Detail screen workflow info box, the Last Updated and Submitted dates were being passed through GFCommon::format_date with a hard-coded 'Y/m/d' date_format parameter. This PR adds a new filter - gravityflow_date_format_entry_detail - to that parameter.

Note for translation / review...The impact is only displayed out of GFCommon::format_date for timestamps older than 24 hours (getting escaped through '%1$s at %2$s') instead of '%s ago'

### DIFF
--- a/class-gravity-flow.php
+++ b/class-gravity-flow.php
@@ -2611,20 +2611,14 @@ PRIMARY KEY  (id)
 			*
 			* @param string $date_format A date format string - defaults to 'Y/m/d'
 			*/
-			$date_created_format = apply_filters( 'gravityflow_date_format_entry_detail', 'Y/m/d' );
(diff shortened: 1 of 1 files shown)


Idealien:
Good catch. Updated in latest commit.

Checking further, it looks like there are only 2 other places where a similar hard-coded 'Y/m/d' format default is used (class-inbox:353, class-status:936). What do you think about renaming the filter to gravityflow_date_format_default and using it in all 4 instances?

stevehenty:
It's a good idea. We could add Gravity_Flow_Common::format_date() and put it there in the future. However, we need to release 2.0.1 so let's go with this hook as it is. Once Richard has approved we can merge it.

PR #98, opened 2018-01-22
5092 status export columns

Idealien:
- Bug Fix - Form column was displaying form id, now showing form title to match status screen.
- Enhancement - Re-uses gravityflow_field_value_status_table filter to display custom columns

### DIFF
--- a/includes/pages/class-status.php
+++ b/includes/pages/class-status.php
@@ -1953,6 +1953,7 @@ class Gravity_Flow_Status_Table extends WP_List_Table {
 		foreach ( $this->items as $item ) {
 			$row_values = array();
 			foreach ( $column_keys as $column_key ) {
+				$col_val = null;
 				if ( array_key_exists( $column_key, $item ) ) {
 					switch ( $column_key ) {
 						case 'created_by' :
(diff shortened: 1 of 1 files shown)


Idealien:
[Status Screen Export - Custom Columns](https://trello.com/c/SUanFhUX/353-status-screen-export-custom-columns)

includes/pages/class-status.php
```diff
@@ -1976,7 +1982,20 @@ public function export() {
...
+					 * Allows the field value to be filtered in the status table.
+					 *
+					 * @since 1.7.1
+					 *
+					 * @param string $label The value to be displayed
+					 * @param int $item['form_id'] The Form ID
+					 * @param string 'id'
+					 * @param array $item The entry array.
+					 */
+					$col_val = apply_filters( 'gravityflow_field_value_status_table', null, $item['form_id'], $column_key, $item );
```
stevehenty: The filter runs for every column on the status table. Is there a reason for adding it here?

Idealien: When viewing the status table in page, the filter runs for every column_key allowing for display of custom columns as well as content overrides of GFlow standard columns. Based on the filter name and params I understood that was intended and have not made any changes to that. 

This PR currently modifies export so that when the status is being exported, the filter only runs for columns which are not a part of the item (entry). Small performance gain, but for consistency should all columns be filtered in export as they are in table display on page?

FYI - The latest update only removes the 'cb' column from ever being considered / filtered.

stevehenty: As the filter only runs inside the else block the filter isn't running for the columns like form_id, created_by, id, workflow_final_status, workflow_step... So the filter would need to be moved outside the else bock to mirror the behaviour on the status table.

Idealien: Filter applies in all column cases now.

PR #105, opened 2018-02-18
Add Step ID # to workflow settings list and edit step screens

Idealien:
**Settings List / Edit Screens show Step ID #**
step_id_display

**How to test:**
- View Workflow Settings screen
- Note Step ID # as new column
- Click on an individual step to edit
- Note the Step ID # is added to sub-heading area (previously just "Step")

### DIFF
--- a/class-gravity-flow.php
+++ b/class-gravity-flow.php
@@ -1160,9 +1160,14 @@ PRIMARY KEY  (id)
 				'choices'    => $step_type_choices,
 			);
 
-
+			if( absint( rgget( 'fid' ) ) !== 0 ) {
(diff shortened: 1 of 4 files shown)


class-gravity-flow.php
```diff
@@ -1162,7 +1162,7 @@ public function feed_settings_fields() {
 
 
 			$settings[] = array(
-				'title'  => 'Step',
+				'title'  => 'Step ID #' . absint( rgget( 'fid' ) ),
```
stevehenty: This needs a check for new IDs. If it's new (zero) then don't display the ID.

stevehenty:
Could you update the screenshot to reflect the updated version, please?

PR #108, opened 2018-02-28
5406 discussion display more

Idealien:
Display more now applies by default on inbox (both edit and display felds) as well as status page.

**Form configuration:**

Add two Discussion type fields
Configure a Workflow user input step with save progress active
Configure an Approval step with the discussion field displayed

**Testing:**

Submit the form.
Access the inbox and save several discussion comments to trigger view more.
(Alternatively consider gravityflow_discussion_items_display_limit to set lower display)
Access the status page and note that view more is now accessible
Access print by browser File menu and GFlow button and note the full list still displays

### DIFF
--- a/includes/fields/class-field-discussion.php
+++ b/includes/fields/class-field-discussion.php
@@ -305,10 +305,6 @@ class Gravity_Flow_Field_Discussion extends GF_Field_Textarea {
 				} else {
 					$return .= $display_items;
 				}
-				$return .= '<style type="text/css" media="print">';
-				$return .= '.gravityflow-dicussion-item-hidden { display:block !important; }';
-				$return .= '.gravityflow-dicussion-item-toggle-display { display: none; }';
-				$return .= '</style>';
 			} else {
 				$return .= $hidden_items . $display_items;
 			}
--- a/includes/fields/class-field-discussion.php
+++ b/includes/fields/class-field-discussion.php
@@ -8,7 +8,7 @@
  */
 
 if ( ! class_exists( 'GFForms' ) ) {
-    die();
+	die();
 }
 
 /**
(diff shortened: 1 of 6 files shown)


js/entry-detail.js
```diff
@@ -5,19 +5,16 @@
...
-        var $toggle = $( '#field_' + formId + '_' + fieldId );
-
-        if ( $toggle ) {
-
-            $toggle.children( '.gravityflow-dicussion-item-hidden' ).slideToggle( 'fast' );
+            $toggle = $(':focus').parent();
+            
+            $toggle.find( '.gravityflow-dicussion-item-hidden' ).slideToggle( 'fast' );
 
             var oldText = $toggle.children( '.gravityflow-dicussion-item-toggle-display' ).attr( 'title' );
```
stevehenty: There are a lot of repeated selectors here. The performance hit is probably not noticeable but, as a rule, we should aim to reuse selectors when possible. They could be stored in a variable and accessed something like this:
var $viewMore = $toggle.children( '.gravityflow-dicussion-item-toggle-display' );
var oldText = $viewMore.attr( 'title' );

includes/fields/class-field-discussion.php
```diff
@@ -252,10 +252,19 @@ public function format_discussion_value( $value, $format = 'html', $entry_id = n
...
 			$hidden_items          = '';
 
-			if ( $entry_id && ! $this->is_form_editor() ) {
+			/**
+			 * Whether to show / hide the toggle to display more discussion items.
+			 *
+			 * @param boolean                       $hide_toggle Whether to prevent the display more toggle from displaying.
+			 * @param Gravity_Flow_Field_Discussion $this        The field currently being processed.
+			 *
+			 * @since 2.0.2.dev
```
stevehenty: The version number would be either 2.0.2 or 2.0.2-dev and should be above the params

PR #110, opened 2018-03-03
Adding filter gravityflow_limit_activity_events

Idealien:
**To Test:**
Add filter to gravityflow_limit_activity_events which returns int value 25
Notice that the Activity page (wp-admin/admin.php?page=gravityflow-activity) returns 25 results.

### DIFF
--- a/includes/pages/class-activity.php
+++ b/includes/pages/class-activity.php
@@ -38,7 +38,17 @@ class Gravity_Flow_Activity_List {
 			return;
 		}
 
-		$events = Gravity_Flow_Activity::get_events();
+		/**
+		*
+		* @since 2.0.2
+		*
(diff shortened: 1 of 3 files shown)


includes/pages/class-activity.php
```diff
@@ -38,7 +38,17 @@ public static function display( $args ) {
...
-		$events = Gravity_Flow_Activity::get_events();
+		/**
+		*
+		* @since 2.0.2
+		*
+		* Allows the limit for activity events to be modified before events are displayed.
+		*
+		* @param int $limit The limit of events.
+		*/
+		$limit = (int) apply_filters( 'gravityflow_limit_activity_events', 400 );
```
stevehenty: According to the naming conventions, this should be gravitylow_event_limit_avtivity_page.

PR #116, opened 2018-03-22
Added gravityflow_step_status_evaluation_approval

Idealien:
***Test  Instructions***

* Create an approval step with more than 2 assignees and approval policy set to 'all'
* Add [this gist](https://gist.github.com/Idealien/8c97b322370b0252ded58d5932f6b050) to your functions.php
* Update the step ID in filter to match your test step
* Test various approval scenarios

The gist is set so that any 2 approvals will complete the step, with any rejection continuing to reject the entire step. Other use cases could apply much different logic such as % of approve/reject with minimum number of non-pending responses.

### DIFF
--- a/includes/steps/class-step-approval.php
+++ b/includes/steps/class-step-approval.php
@@ -384,9 +384,9 @@ class Gravity_Flow_Step_Approval extends Gravity_Flow_Step {
 		 *
 		 * @since 2.1-dev 
 		 *
-		 * @param string                $step_status   The status of the step
-		 * @param array                 $assignees     Array of Gravity_Flow_Assignee objects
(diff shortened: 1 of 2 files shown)


includes/steps/class-step-approval.php
```diff
@@ -379,6 +379,17 @@ public function status_evaluation() {
 			}
 		}
 
+		/**
+		 * Allows the step status for the approval to be customized
+		 *
+		 * @since 2.1-dev 
+		 *
+		 * @param string                $step_status   The status of the step
+		 * @param array                 $assignees     Array of Gravity_Flow_Assignee objects
```
stevehenty: This can be `@param Gravity_Flow_Assignee[]`

PR #118, opened 2018-03-25
Revise export  to output pending status with duration

Idealien:
***Test Instructions***
- Use master branch and go to status page
- Note that status column for pending entries includes the duration since step was started
- Perform export and note that status column for pending entries only includes the status.
- Switch to branch 5436-status-export-pending
- Note that both status table and export contain the duration (export has it as separate column)


Idealien:
Revise export function to include duration column to bring consistency to info shown on status table and export. Value will only be populated when workflow_final_status column == pending.

PR #124, opened 2018-04-07
Fix status update for recent assignee refactor

stevehenty:
The recent assignee refactor works only for the User Input step. This PR adds support also for the Approval step.

**Testing instructions**
On master
- add the following snippet to add a dept assignee:
https://gist.github.com/stevehenty/dc73ca8c9501d11f28419ce92399aa3b
- add the department as an assignee of an approval step
- submit the form and reach the approval step
- when approving or rejecting the step status doesn't change  making it impossible to complete the step

On this branch
- the approval step works correctly both from the workflow detail page and from the quick actions column on the inbox

### DIFF
--- a/includes/steps/class-step.php
+++ b/includes/steps/class-step.php
@@ -296,7 +296,6 @@ abstract class Gravity_Flow_Step extends stdClass {
 		foreach ( $assignees as $assignee ) {
 			if ( $assignee->is_current_user() ) {
 				return true;
-				break;
 			}
 		}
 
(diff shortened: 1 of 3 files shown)


includes/steps/class-step.php
```diff
@@ -296,8 +291,13 @@ public function rest_permission_callback( $request ) {
...
 		}
 
-		if ( $this->is_assignee( $assignee_key ) ) {
-			return true;
+		$assignees = $this->get_assignees();
+
+		foreach ( $assignees as $assignee ) {
+			if ( $assignee->is_current_user() ) {
+				return true;
+				break;
```
richardW8k: break can be removed as it will never be reached

PR #128, opened 2018-04-29
Add the "All fields except" choice to the Display Fields setting

richardW8k:
This adds a new choice to the Display Fields setting allowing all fields except the selected fields to be displayed on the workflow detail page.

**Testing Instructions**

Check approval and user input steps which use the "all fields" and "selected fields" choices continue to display the same fields as master.

Configure approval and user input steps with the "all fields except" choice selected and one or more fields selected which shouldn't be displayed. Submit the form and then progress through the workflow checking the selected fields are not displayed.


stevehenty:
Can we make the labels a bit more explicit? Instead of:
```
All Fields
All fields Except
Selected fields
```
How about this?
```
Display all fields
Display all fields except selected
Hide all fields except selected
```

PR #129, opened 2018-05-09
Fix one-click actions

mgmartel:
The one-click actions (approve and reject) from emails weren't working anymore because the tokens were created without an action. This PR fixed that.


stevehenty:
Thanks @mgmartel !

PR #130, opened 2018-05-11
Load users only when needed

stevehenty:
This PR fixes a performance issue which can affect sites with a lot of users. The list of users is currently generated on every admin request. The WordPress function `get_users()` performs one query for every user meaning that a site with 1000 users will generate 1000 unnecessary requests on every admin page load. This PR generates the list only on the settings page.
Fixes [HS5993](https://secure.helpscout.net/conversation/578178309/5993/?folderId=516731)

### DIFF
--- a/class-gravity-flow.php
+++ b/class-gravity-flow.php
@@ -503,7 +503,7 @@ PRIMARY KEY  (id)
 				}
 			}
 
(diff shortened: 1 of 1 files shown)


cbrcongit:
Reading the diff view on this was a pleasure. Simply insightful, simply.

PR #138, opened 2018-05-21
Updates filter_gform_pre_render() method to only apply to current form.

spivurno:
In some situations (like with GP Nested Forms), the gform_pre_render may be called multiple times in a single page load. Currently, Gravity Flow will filter all forms that are filtered through gform_pre_render when editing an entry.


stevehenty:
Thanks @spivurno !

PR #143, opened 2018-06-03
Fix RTL entry detail

stevehenty:
Fixes an issue on some RTL themes where the content is too narrow. e.g. on TwentySeventeen
rtl-issue
Fixes [HS6133](https://secure.helpscout.net/conversation/593320004/6133?folderId=1113418)


stevehenty:
closed in favour of #144

PR #146, opened 2018-06-29
Fix status shortcode attributes

stevehenty:
This PR fixes an issue with the status page shortcode where the workflow_info and step_status attributes are ignored.


richardW8k:
The attributes are now being used but if both are set to false the empty postbox is still displayed: https://www.dropbox.com/s/4z2t6iwjtg868o0/Screenshot%202018-06-30%2013.55.55.png?dl=0

PR #148, opened 2018-07-01
HS#5583 - Re-factor get_value_export for multiple user field types

Idealien:
A resolution to https://github.com/gravityflow/gravityflowformconnector/pull/6 which is now closed.

The form export attached has two forms that validate the form connector assignment by field works for the GFlow field types (assignee - by user, role, user, role and multi-user). They all also work for export. PR also passes all unit and acceptance tests currently.

[gravityforms-export-2018-07-01.json.zip](https://github.com/gravityflow/gravityflow/files/2153269/gravityforms-export-2018-07-01.json.zip)

Adding Steven and Richard for review


Idealien:
[gravityforms-export-2018-07-01.json.zip](https://github.com/gravityflow/gravityflow/files/2153269/gravityforms-export-2018-07-01.json.zip)

PR #160, opened 2018-09-01
Outgoing webhook response mapping for JSON values 

Idealien:
**Enhancement**
Outgoing webhook steps now have option to map field values from the webhook response.
image

Automated Test Case 0027 includes 3 scenarios:
- No mapping (existing functionality)
- Mapping base field types with values from response body (JSON)
- Negative testing that field mapping for a non-JSON response body does not fail

Test outgoing URL responses are provided by stub. See tests/acceptance-tests/mu-plugins/webhook-response-stubs.php that filters pre_http_request for any attempted requests to unit-test-webhook.com 

**Additional Hook(s)**
Filter: gravityflow_entry_webhook_response_mapping
Fires during the mapping of response values providing a way to do targeted adjustments to the response data when mapping into fields. Next best alternative would be a larger loop and custom code relying on gravityflow_post_webhook action after all the timeline and response code info was stored.

**Example response mapping (and filter usage) with a public API**
[Form Export](https://github.com/gravityflow/gravityflow/files/2363635/gravityforms-export-2018-09-08.json.zip)
An outgoing step performs a trademark lookup via https://www.markerapi.com/ - API requires free sign-up for user/pass that is added to request URL (see REPLACEME values in request path).
Without filter, the response mapping only results in the 'count' field showing how many trademarks match the selected term. With filter ([and this gist](https://gist.github.com/Idealien/02788505327ca36cca0fde0b3ec3ce1f)) any matched trademarks will have wordmark, code and description written into the paragraph field.


includes/steps/class-step-webhook.php
```diff
@@ -624,6 +662,43 @@ function send_webhook() {
...
+
+				if ( ! is_array( $this->response_mappings ) ) {
+					$this->log_debug( __METHOD__ . '(): Step has no response mappings defined' );
+					return;
+				}
+
+				$data = json_decode( wp_remote_retrieve_body( $response ), TRUE );
+
+				//PHP >= 5.3?
+				if ( json_last_error() !== 0 ) {
```
stevehenty: We'll need to check for the existence of the function here and provide a fallback method for 5.2.

tests/acceptance-tests/_mu-plugins/webhook-response-stubs.php
```diff
@@ -0,0 +1,139 @@
+<?php
+/**
+Plugin Name: Webhook Response Stubs
+Plugin URI: https://gravityflow.io
+Description: Stub outgoing webhook requests and provide expected response code/body for tests
+Version: 1.0
+Author: Gravity Flow
+Author URI: https://gravityflow.io
+License: GPL-2.0+
```
stevehenty: We ought to ensure that all files have a copyright notice.

PR #165, opened 2018-09-09
Add invalid license admin notice

stevehenty:
This PR as an admin notice when the license is not valid. It also hides the license key setting when the GRAVITY_FLOW_LICENSE_KEY constant is defined.

**Testing Instructions**
- Enter an invalid key and the admin notice will appear.
- Dismiss the notice and it will not appear for the rest of the day.
- The next day the admin notice will appear again and can be dismissed until the following day.
- The invalid license message will appear on every admin page in the site.
- The "Active License" and "get a license" buttons will appear with the message as long as the current page is not the settings page.
- Defining the GRAVITY_FLOW_LICENSE_KEY constant will hide the license key setting but the invalid license message will be displayed when the transient with the license details expires.

### DIFF
--- a/class-gravity-flow.php
+++ b/class-gravity-flow.php
@@ -244,6 +244,8 @@ if ( class_exists( 'GFForms' ) ) {
 			if ( $this->is_app_settings() ) {
 				require_once( GFCommon::get_base_path() . '/tooltips.php' );
 			}
+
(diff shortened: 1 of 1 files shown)


stevehenty:
@richardW8k good catch, thanks. Should be fixed now.

PR #168, opened 2018-09-23
Add support for parsing the response

stevehenty:
The PR adds support for parsing complex webhook response values. It also adds the `gravityflow_webhook_url_in_note` filter

For example:

For the response:

`{ "translations": [ { "detected_source_language": "DE", "text": "Hello World!" } ] }`

The following key will extract the value of the value `text` key ("Hello World!"):

`translations\0\text`

Testing instructions
Unzip and import the test form:
[zippopotam-sample-form2.json.zip](https://github.com/gravityflow/gravityflow/files/2408552/zippopotam-sample-form2.json.zip)
Preview the form, select US and enter 23464 in the zip code field and submit. The place name longitude and latitude fields will be populated.
zipcode-example

### DIFF
--- a/includes/steps/class-step-webhook.php
+++ b/includes/steps/class-step-webhook.php
@@ -812,7 +812,7 @@ class Gravity_Flow_Step_Webhook extends Gravity_Flow_Step {
 
 		$target_field = GFFormsModel::get_field( $form, $target_field_id );
 
-		if ( $target_field instanceof GF_FIELD ) {
+		if ( $target_field instanceof GF_Field ) {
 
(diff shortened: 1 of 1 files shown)


includes/steps/class-step-webhook.php
```diff
@@ -881,6 +901,30 @@ public function add_mapping_to_entry( $entry, $mapping, $data, $content_type ) {
 		return $entry;
 	}
 
+	public function parse_response_value( $value, $key, $default = '' ) {
```
richardW8k: Just needs a docblock

PR #173, opened 2018-10-08
Fix admin notice style

yoren:
This PR adds the `<p>` tag to the admin screen feedback when it's missing in the original message.

### Admin notice before this PR
screenshot 2018-10-03 17 03 23

### Admin notice after this PR
screenshot 2018-10-03 17 05 10

### DIFF
--- a/class-gravity-flow.php
+++ b/class-gravity-flow.php
@@ -4309,7 +4309,7 @@ PRIMARY KEY  (id)
 					}
 					?>
 					<div class="notice error is-dismissible gravityflow_validation_error" style="padding:6px;">
-						<?php echo esc_html( $feedback->get_error_message() ); ?>
+						<p><?php echo esc_html( $feedback->get_error_message() ); ?></p>
 					</div>
 					<?php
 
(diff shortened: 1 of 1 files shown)


class-gravity-flow.php
```diff
@@ -4318,6 +4318,9 @@ public function inbox_page( $args = array() ) {
 
 					$entry = GFAPI::get_entry( $entry_id ); // Refresh entry.
 
+					if ( substr( $feedback, 0, 3 ) !== '<p>' ) {
```
yoren: @stevehenty Instead of adding the `<p>` tag to feedback messages everywhere, I feel it's easier to just do a check and add it before rendering. Will there be any side effect of this approach?

stevehenty: I agree it's certainly quicker and easier. It's also a bit less clear what's going on and we can't guarantee that all messages will need to start with a `p` tag in the future. We may decide one day to add a message starting with a different tag or perhaps a customer has a filter which is styled in a specific way so I think it's best to add the tag to the markup. If the markup is repetitive then we should look into creating a method for it.

yoren: @stevehenty ok I've removed it. Now my question would be, we usually provide a filter for feedback like this: https://github.com/gravityflow/gravityflow/blob/master/class-gravity-flow.php#L4659

Is it okay if I add the `<p>` tags to all `$feedback` messages before this filter? Which means the `$feedback` will contain HTML tags and it may not be ideal because people may expect it's a text string?

stevehenty: Yeah, it's getting complicated. Let's revert back to your original idea of checking for the `p` tag.

yoren:
@stevehenty I just reverted. Let me know if that works for us now.

PR #175, opened 2018-10-10
Approval Step Revert Notification + User Input Assignee Notification Override

Idealien:
**This feature PR:**

1. Enhances the Approval Step Notifications to include a new tab for Revert message.
It is only visible when the Revert setting has been selected.
2. When Revert notification is active, by default the user input assignee notification is disabled.
3. Adds the gravityflow_user_input_notification_send_on_revert filter to allow user input notification to continue to be sent when the revert notification has been defined / sent.
4. Adds acceptance test (0028) which validates approval, revert and notifications sent

**To test this feature:**
- Execute automated test form/flow 0028 or follow manual step instructions below.
- Create a form which has an approval step followed by user input step.
- Activate the 'Revert to User Input step' option.
- See that the Revert Email tab in notifications is available.
- Define a Revert Email and save settings.
- Submit the form and revert the approval step during workflow.

If you add the following code snippet to functions.php and repeat the test procedure, both the user input assignee notification and the revert notification will be sent.

```
add_action(` 'gravityflow_user_input_notification_send_on_revert', 'sh_revert_override', 10, 5);
  function sh_revert_override( $override, $notification, $form, $entry, $step ) 
    return true;
  }
```


includes/steps/class-step-approval.php
```diff
@@ -1032,6 +1065,41 @@ public function get_revert_icon() {
 		$revert_icon  = '<i class="fa fa-undo" style="color:blue"></i>';
 		return $revert_icon;
 	}
+
+	/**
+	 * Ensure User Input assignee notification does not send if an approval revert notification exists with override selected
+	 *
+	 * @since 2.3.2-dev
```
stevehenty: Let's use since 2.3.2

includes/steps/class-step-approval.php
```diff
@@ -1032,6 +1065,41 @@ public function get_revert_icon() {
...
+	 *
+	 * @param string                $notification     The potential notification
+	 * @param array                 $form             The current form array.
+	 * @param array                 $entry            The current entry.
+	 * @param Gravity_Flow_Step     $step             The current step
+	 */
+	public static function gravityflow_user_input_notification_send_on_revert( $notification, $form, $entry, $step ) {
```
stevehenty: By convention, we tend to use the name of the hook prefixed by the type. So the name of the method would be `filter_gravityflow_notification`.

includes/steps/class-step-approval.php
```diff
@@ -1032,6 +1065,41 @@ public function get_revert_icon() {
...
+		if ( $step->get_type() == 'user_input' ) {
```
stevehenty: I can't think of a scenario which requires it right now but it might be safer, more future proof, to check the step ID also.

includes/steps/class-step-approval.php
```diff
@@ -1032,6 +1065,41 @@ public function get_revert_icon() {
...
+			 *
+			 * @param bool                 $override      Whether the user input assignee notification should be sent
+			 * @param array                $notification  The notification object which would be sent for user input assignee
+			 * @param array                $form          The current form.
+			 * @param array                $entry         The current entry.
+			 * @param Gravity_Flow_Step    $step          The current step
+			 *
+			 * return bool
+			 */
+			$override = apply_filters( 'gravityflow_user_input_notification_send_on_revert', false, $notification, $form, $entry, $step );
```
stevehenty: Can we omit the filter until we get a request which requires it?

includes/steps/class-step-approval.php
```diff
@@ -300,8 +302,23 @@ public function get_settings() {
...
 			);
 			$settings['fields'][]           = $revert_field;
+
+			$notification_tabs['tabs'][] = array(
+				'label'  => __( 'Revert Email', 'gravityflow' ),
+				'id'     => 'tab_revert_notification',
+				'fields' => $settings_api->get_setting_notification( array(
+					'name_prefix'      => 'revert',
+					'checkbox_label'   => __( 'Send email when the entry is reverted to a user input step', 'gravityflow' ),
+					'checkbox_tooltip' => __( 'Enable this setting to send an email when the entry is reverted to a user input.', 'gravityflow' ),
```
stevehenty: Let's add a note clarifying that the assignee email of the user input step will not be sent if this setting is enabled.

includes/steps/class-step-approval.php
Idealien: Update now includes a check between current user input step ID and the approval revert step selected.

richardW8k:
The acceptance test and associated form will need renumbering as 0028 already exists in #169 and 0029-0033 exist in the add-merge-tag-tests branch.

Idealien:
> The acceptance test and associated form will need renumbering as 0028 already exists in #169 and 0029-0033 exist in the add-merge-tag-tests branch.

Updated to 0034

PR #177, opened 2018-10-13
Update date/time to respect WP general settings

Idealien:
Refer to HS#6938
The current date/time format applied to each note in the timeline is based on the default of Gravity_Flow_Common::format_date which is "d M Y g:i a". This PR refines that to set the format based on the WP options date_format and time_format.

**To Test:**

1. View any inbox or status page with timeline enabled when on master branch.
1. Checkout this branch and update your WP date/time settings on general options page.
1. Note that the date/time in upper right corner of each note matches selected settings.


stevehenty:
Updated to address all dates

PR #182, opened 2018-10-19
Ensure the PDF methods exist before trying to use it

jakejackson1:
Resolve a fatal error when Gravity PDF is activated, but the system requirements aren't met, and a Gravity Flow step is loaded.


stevehenty:
Thanks @jakejackson1

PR #187, opened 2018-10-28
Status page - step column display info for queued/expires

Idealien:
The status page step column previously would show both queued and expiration date / time. The expiration date/time was inaccurate as it would not be actually determined until the queue had expired. This fix changes the display logic to only show expiration date/time when it step status is not queued.

Test case with all combinations of queue / expiration
[queue-expire-test.zip](https://github.com/gravityflow/gravityflow/files/2522926/queue-expire-test.zip)


Idealien:
[Status page: Don't display expiration date if step is queued](https://trello.com/c/NR6L0j6O/443-status-page-dont-display-expiration-date-if-step-is-queued)

includes/pages/class-status.php
```diff
@@ -974,9 +974,7 @@ public function column_workflow_step( $item ) {
 
 				$label .= '' . sprintf( '<div class="step_status_queue">(%s: %s)</div>', esc_html__( 'Queued', 'gravityflow' ), $scheduled_date );
 
-			}
-
-			if ( $step->supports_expiration() && $step->expiration ) {
+			} else if ( $step->supports_expiration() && $step->expiration ) {
```
stevehenty: "[Use `elseif`, not `else if`](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#use-elseif-not-else-if)"

PR #188, opened 2018-11-02
Allow multiple Gravity PDFs to be sent with a single notification

jakejackson1:
Still a WIP. I want to more thoroughly test both the upgrade routine and the multi-PDF sending code.

Let me know if there's a better way to access the PDF checkbox settings for a step.

### DIFF
--- a/includes/steps/class-step.php
+++ b/includes/steps/class-step.php
@@ -957,10 +957,14 @@ abstract class Gravity_Flow_Step extends stdClass {
 			$notification['subject'] = $this->{$subject};
 		}
 
-		if ( defined( 'PDF_EXTENDED_VERSION' ) && version_compare( PDF_EXTENDED_VERSION, '4.0-RC2', '>=' ) ) {
-			if ( $this->{$type . 'gpdfEnable'} ) {
-				$gpdf_id      = $this->{$type . 'gpdfValue'};
-				$notification = $this->gpdf_add_notification_attachment( $notification, $gpdf_id );
+		/* Attach PDF(s) if required */
+		if ( method_exists( 'GPDFAPI', 'get_form_pdfs' ) ) {
+			$form_pdfs = GPDFAPI::get_form_pdfs( $this->get_form_id() );
+			foreach ( $form_pdfs as $fprm_pdf ) {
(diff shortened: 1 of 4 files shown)


jakejackson1:
@stevehenty No problem. I've completed my manual testing and both the upgrade code and the multi-PDF attachment code are sound. 

I noted you added this to the 2.4 milestone so I've updated the upgrade code naming / version to match.

Side note: In the Notification Step, the PDF field isn't shown / hidden when the Workflow Notification fields is ticked. Is there any easy way to fix that? It's a bit confusion showing the PDF setting when the option is disabled.

includes/steps/class-step.php
```diff
@@ -957,9 +957,11 @@ public function get_notification( $type ) {
 			$notification['subject'] = $this->{$subject};
 		}
 
-		if ( defined( 'PDF_EXTENDED_VERSION' ) && version_compare( PDF_EXTENDED_VERSION, '4.0-RC2', '>=' ) ) {
-			if ( $this->{$type . 'gpdfEnable'} ) {
-				$gpdf_id      = $this->{$type . 'gpdfValue'};
+		/* Attach PDF(s) if required */
```
stevehenty: How about something like this:
```
if ( method_exists( 'GPDFAPI', 'get_form_pdfs' ) ) {
	$form_pdfs = GPDFAPI::get_form_pdfs( $this->get_form_id() );
	foreach ( $form_pdfs as $fprm_pdf ) {
		$pdf_key = $type . 'gravitypdf_' . $fprm_pdf['id'];
		if ( $this->{$pdf_key} ) {
			$notification = $this->gpdf_add_notification_attachment( $notification, $fprm_pdf['id'] );
		}
	}
}
```

stevehenty:
To hide/show the pdf setting on the notification step you should be able to just add `workflow_notification_gpdf` to both the lists of settings in `toggleWorkflowNotificationType()`. You'll need to add it to the list of sub settings in `setSubSettings()` below `workflow_notification_autoformat`.

jakejackson1:
@stevehenty both suggestions worked perfectly. The PR has been updated appropriately. Thanks for that.

PR #194, opened 2018-11-17
Add support for Repeater fields

stevehenty:
This PR adds support to the User Input step for Repeater fields (added in Gravity Forms 2.4) editable fields. It also batches multiple field updates into single queries to improve performance - this bumps the minimum version Gravity Forms requirement 2.3.

**Testing in instructions**
- Ensure the Repeater field works as expected as display and editable fields on Gravity Forms 2.4.
- Check that no regression issues are introduced when using Gravity Forms 2.3.

### DIFF
--- a/includes/pages/class-entry-editor.php
+++ b/includes/pages/class-entry-editor.php
@@ -598,6 +598,10 @@ class Gravity_Flow_Entry_Editor {
 			$display_value = GFCommon::get_lead_field_display( $field, $value, $this->entry['currency'] );
 		}
 
+		if ( ! empty( $field->fields ) ) {
+			$html .= sprintf( '<label class="gfield_label">%s</label>', $field->label );
+		}
+
 		$display_value = apply_filters( 'gform_entry_field_value', $display_value, $field, $this->entry, $this->form );
 
 		if ( $this->display_empty_fields ) {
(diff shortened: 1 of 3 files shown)


richardW8k:
If the repeater is not selected as editable but is a display field the field label is not being displayed.

screenshot 2018-11-24 08 46 29

richardW8k:
Child fields which are required are being validated when saving progress which is inconsistent with the recent change introduced in #183.

includes/pages/class-entry-editor.php
```diff
@@ -420,7 +420,15 @@ public function filter_gform_field_input( $html, $field ) {
 		$value = $this->get_post_image_value( $value, $field );
 		$value = $this->get_post_category_value( $value, $field );
 
-		$html = $field->get_field_input( $this->form, $value, $this->entry );
+		if ( $this->step instanceof Gravity_Flow_Step_User_Input && ! empty( $field->fields ) &&  rgpost( 'gravityflow_status' ) == 'in_progress') {
```
richardW8k: Is this required? I would have thought changing the property in `Gravity_Flow_Step_User_Input::validate_editable_fields()` would be enough and that removing the required icon on render could cause some confusion.

stevehenty: Without it, the validation message is displayed making it look like there was an error.

stevehenty: I agree it's confusing to remove the required field indicator but I couldn't find a better way to handle it.

PR #199, opened 2019-01-03
Add back link to shortcode

Idealien:
This PR enables a back link to be added to the entry detail screen to return user to the inbox/status page based on shortcode. 

### Shortcode Attributes
- **back_link** - Default is false. Set to true to show the back link on entry detail page
- **back_link_text** - Default is 'Return to list'. Set a custom message per shortcode.
- **back_link_url** - Default is null. Enable non-GF shortcode pages to set a return link to their summary view.

### Filter(s)
- **gravityflow_back_link_url_entry_detail** - To allow customization of the link via code. The shortcode $args are passed as the 2nd argument allowing devs to control which scenarios they may want to customize.

### Other
- Updates the Gravity View integration of workflow_detail_link field. If the default GFlow inbox/shortcode page has the back_link=true attribute to its' shortcode, the back link will return to the current view page instead.

The shortcode attributes and filter will need documentation updates as a part of the version launch.

### DIFF
--- a/includes/class-gravityview-detail-link.php
+++ b/includes/class-gravityview-detail-link.php
@@ -74,7 +74,7 @@ class Gravity_Flow_GravityView_Workflow_Detail_Link extends GravityView_Field {
 	private function add_hooks() {
 		add_filter( 'gravityview_entry_default_fields', array( $this, 'add_entry_default_field' ), 10, 3 );
 		add_filter( 'gravityview_field_entry_value_workflow_detail_link', array( $this, 'modify_entry_value_workflow_detail_link' ), 10, 4 );
-		add_filter( 'gravityflow_back_link_entry_detail', array( $this, 'modify_entry_detail_back_link' ), 10, 2 );
+		add_filter( 'gravityflow_back_link_url_entry_detail', array( $this, 'modify_entry_detail_back_link' ), 10, 2 );
 	}
 
 	/**
--- a/includes/class-gravityview-detail-link.php
+++ b/includes/class-gravityview-detail-link.php
@@ -132,7 +132,7 @@ class Gravity_Flow_GravityView_Workflow_Detail_Link extends GravityView_Field {
 
 		$url = Gravity_Flow_Common::get_workflow_url( $query_args, $page_id );
 
-		if( $page_id != 'admin' ) {
+		if ( $page_id != 'admin' ) {
 			global $post;
 			$url .= '&gvp=' . $post->ID;
 		}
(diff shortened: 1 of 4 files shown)


includes/class-gravityview-detail-link.php
```diff
@@ -131,6 +132,11 @@ function modify_entry_value_workflow_detail_link( $output, $entry, $field_settin
 
 		$url = Gravity_Flow_Common::get_workflow_url( $query_args, $page_id );
 
+		if( $page_id != 'admin' ) {
```
stevehenty: Missing a space after the `ìf`

includes/pages/class-entry-detail.php
```diff
@@ -269,6 +273,44 @@ class="gf_admin_page_formid">ID: <?php echo absint( $form['id'] ); ?></span>
 		<?php
 	}
 
+	/**
+	 * Displays the back link on entry detail page if enabled.
+	 *
+	 * @since 2.5-dev
```
stevehenty: @since 2.5 is fine for all the since tags

includes/class-gravityview-detail-link.php
```diff
@@ -169,6 +175,37 @@ function field_options( $field_options, $template_id, $field_id, $context, $inpu
...
+	 *
+	 * @param string $url       The potential back link URL.
+	 * @param array  $args      The shortcode arguments for the page where the back link is to be displayed
+	 *
+	 * @since 2.5-dev
+	 *
+	 * @return string
+	 */
+	function modify_entry_detail_back_link( $url, $args ) {
+		//Verify the back link is set as active on current page and the potential back link has querystring params
```
stevehenty: Missing space after // for the inline comments.

includes/pages/class-entry-detail.php
```diff
@@ -269,6 +273,44 @@ class="gf_admin_page_formid">ID: <?php echo absint( $form['id'] ); ?></span>
...
+		 * Useful in cases where the access into entry detail page is not based out of gravityflow shortcode.
+		 *
+		 * @since 2.5
+		 *
+		 * @var string $url    The customized URL to redirect user to when clicking the back link
+		 * @var array  $args   The shortcode attributes for the current page
+		 *
+		 * @return string
+		 */
+		$url = apply_filters( 'gravityflow_back_link_entry_detail', $url, $args );
```
stevehenty: `gravityflow_back_link_url_entry_detail`would allow us to add a `gravityflow_back_link_entry_detail`later if we need to.

PR #201, opened 2019-01-06
Add update user step

stevehenty:
Adds the Update User Step which updates user meta and properties.

It also removes the generic_map setting as it's now in the add-on framework.

It also adds PHP7.3 to the Travis CI test matrix.

**Testing Instructions**
- Add an Update User Step and map the properties and meta keys to fields or custom values and check that all properties, roles and user meta are set correctly.
- Check each of the different options for identifying the user work as expected.
- The custom meta value supports merge tags and shortcodes e.g. [gv_math scope={entry_id}]{user:rank}+1[/gv_math]
- Unit tests included - check for completeness.
- Check for regressions as a result of switching to the framework version of the generic_map setting.


includes/steps/class-step-update-user.php
```diff
@@ -0,0 +1,687 @@
...
+							'value' => 'replace',
+						),
+						array(
+							'label' => esc_html__( 'Add to existing roles', 'gravityflow' ),
+							'value' => 'add',
+						),
+					),
+					'onchange' => 'jQuery(this).closest("form").submit();',
+					'class'    => 'medium',
+					'tooltip'  => sprintf( '<h6>%s</h6> %s', esc_html__( 'Nickname', 'gravityflow' ), esc_html__( 'Select the form field that should be used for the user\'s roles.', 'gravityflow' ) ),
```
richardW8k:
```suggestion
					'tooltip'  => sprintf( '<h6>%s</h6> %s', esc_html__( 'Roles', 'gravityflow' ), esc_html__( 'Select how the user\'s roles should be updated.', 'gravityflow' ) ),
```

includes/steps/class-step-update-user.php
```diff
@@ -0,0 +1,687 @@
...
+					$value = do_shortcode( $value );
+				} else {
+					$value = gravity_flow()->get_field_value( $form, $entry, $meta_value );
+
+					if ( $wc_active && in_array( $meta_key, array(
+							'billing_state',
+							'shipping_state',
+						) ) ) {
+						$field = GFFormsModel::get_field( $form, $meta_value );
+						if ( $field->get_input_type() == 'address' && $meta_value = $field->formId . '.4' ) {
```
richardW8k: Check we have a field, just in case the field is deleted and they forget to update the step.
```suggestion
						if ( $field && $field->get_input_type() == 'address' && $meta_value = $field->formId . '.4' ) {
```

includes/steps/class-step-update-user.php
```diff
@@ -0,0 +1,687 @@
...
+							// Update the state name to the state code after the country has been updated.
+							$wc_meta[ $meta_key ] = $value;
+						}
+					}
+
+					if ( in_array( $meta_key, array( 'billing_country', 'shipping_country' ) ) ) {
+						$field = GFFormsModel::get_field( $form, $meta_value );
+						if ( $field->get_input_type() == 'address' && $meta_value = $field->formId . '.6' ) {
```
richardW8k:
```suggestion
						if ( $field && $field->get_input_type() == 'address' && $meta_value = $field->formId . '.6' ) {
```

includes/steps/class-step-update-user.php
```diff
@@ -0,0 +1,687 @@
...
+							'shipping_state',
+						) ) ) {
+						$field = GFFormsModel::get_field( $form, $meta_value );
+						if ( $field && $field->get_input_type() == 'address' && $meta_value = $field->formId . '.4' ) {
+							// Update the state name to the state code after the country has been updated.
+							$wc_meta[ $meta_key ] = $value;
+						}
+					}
+
+					if ( in_array( $meta_key, array( 'billing_country', 'shipping_country' ) ) ) {
```
richardW8k: Should this also check `$wc_active`?

includes/steps/class-step-update-user.php
```diff
@@ -0,0 +1,687 @@
...
+			case 'select_user_field':
+				$entry = $this->get_entry();
+				if ( $entry[ $this->user_field ] ) {
+					$user = get_user_by( 'ID', $entry[ $this->user_field ] );
+				}
+				break;
+			case 'select_email_field':
+				$entry = $this->get_entry();
+				if ( $entry[ $this->email_field ] ) {
+					$user = get_user_by( 'email', $entry[ $this->user_field ] );
```
richardW8k: Should this be using `$this->email_field`?

includes/steps/class-step-update-user.php
```diff
@@ -0,0 +1,687 @@
...
+
+			// Update roles
+
+			if ( $this->roles_action == 'replace' ) {
+				foreach ( $user->roles as $role ) {
+					$user->remove_role( $role );
+				};
+			}
+
+			foreach ( $this->roles as $new_role ) {
```
richardW8k: We should check that `$this->roles` is an array. If no roles are selected on the step you get an "invalid argument supplied for foreach()" warning.

PR #202, opened 2019-01-11
Add filter to gform_is_value_match for multiuser field values

Idealien:
Refer to HS 7547

This PR resolves an issue where conditional routing with a multi-user field could lead to incorrect number of assignees being set for a step. The default GForms value match for 'contains' logic type would return true for potential route of selected User ID of 1 with a multiuser field containing selection of user ID 10.

**Testing Instructions**

- Create a form with a multiuser field
- Ensure your site has 10+ users
- Create an approval step with conditional routing for soft matching user IDs (2 & 22, 7 & 57, etc)
- Submit the form and select the larger user ID by name
- Verify that the step has been assigned to more users than is expected
- Switch to the PR branch and restart the step
- Verify that only the expected users have been set as assignee

### DIFF
--- a/class-gravity-flow.php
+++ b/class-gravity-flow.php
@@ -194,7 +194,7 @@ if ( class_exists( 'GFForms' ) ) {
 			add_filter( 'gform_enqueue_scripts', array( $this, 'filter_gform_enqueue_scripts' ), 10, 2 );
 			add_filter( 'gform_pre_replace_merge_tags', array( $this, 'replace_variables' ), 10, 7 );
 
-			add_filter( 'gform_is_value_match', array( $this, 'filter_value_match_multiuser'), 10, 6 );
+			add_filter( 'gform_is_value_match', array( $this, 'filter_value_match_multiuser' ), 10, 6 );
 
 			add_action( 'gform_entry_created', array( $this, 'action_entry_created' ), 8, 2 );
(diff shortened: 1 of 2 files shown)


class-gravity-flow.php
```diff
@@ -7639,6 +7641,26 @@ public function evaluate_conditional_logic( $logic, $form, $entry ) {
...
+		/**
+		 * Determines if a multiuser field is being used for conditional routing and has an exact value match of user ID
+		 * 
+		 * @since 2.5
+		 *
+		 * @return bool
+		 */
+		public function filter_value_match_multiuser( $is_match, $field_value, $target_value, $operation, $source_field, $rule ) {
+			
+			if ( $source_field->type !== 'workflow_multi_user' ) {
```
richardW8k: It should also return early when the current rule is based on an entry property or meta
```suggestion
			if ( ! $source_field || $source_field->type !== 'workflow_multi_user' ) {
```

class-gravity-flow.php
```diff
@@ -7639,6 +7641,26 @@ public function evaluate_conditional_logic( $logic, $form, $entry ) {
...
+				return $is_match;
+			}
+
+			if( in_array( $target_value, $field_value, true ) ) {
```
richardW8k: Missing one space
```suggestion
			if ( in_array( $target_value, $field_value, true ) ) {
```

PR #204, opened 2019-01-26
Fix inbox excess entries page size limit

richardW8k:
re: [HS#7586](https://secure.helpscout.net/conversation/748963379/7586?folderId=1113535)

Fixes an issue where the inbox page excess entries indicator continues to display the limit as 150 when the page size has been changed using the gravityflow_inbox_paging filter.

## Testing
- Without the filter check that the Gravity Flow default limit of 150 is displayed
- Add the filter setting a custom limit and then check that the limit is updated on the inbox page
  ```
  add_filter( 'gravityflow_inbox_paging', function() {
      return array( 'offset' => 0, 'page_size' => 30 );
  } ); 
  ```
- Add the filter without the page_size arg and then check that a limit of 20 is displayed which matches the default page size used by GFAPI::get_entries()
  ```
  add_filter( 'gravityflow_inbox_paging', function() {
      return array( 'offset' => 0 );
  } ); 
  ```

### DIFF
--- a/change_log.txt
+++ b/change_log.txt
@@ -1,3 +1,4 @@
 - Fixed the Gravity Perks Nested Forms Add-On integration not delaying the workflow for child entries created before the parent form is submitted.
 - Fixed an issue where Conditional Routing based on what values a Multiuser field contains could lead to extra users being set as step assignees.
 - Fixed conditional assignee routing using the post category field.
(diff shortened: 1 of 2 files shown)


Idealien:
The PR functionality is good.
Just missing a change_log.txt note to feed release notes for the bug fix.

PR #208, opened 2019-02-03
Add start and complete settings

stevehenty:
Adds Start and Complete settings to the workflow.

The Start settings include a special workflow conditional logic setting to decide whether to process the workflow or just complete immediately.

Both the Start and Complete settings have:
- The instructions setting especially for form submitters for while their workflow is pending and complete.
- The display fields setting for form submitters. (Users with View All caps can see all fields)

It's only possible to add one Start step and one Complete step to each workflow. The Start step always appears at the top of the list and the Complete step always appears at the bottom.

Both the start and complete settings are added automatically when the first step is added to new workflows. Existing workflows can add the steps manually from the step list. When the last step is removed the Start and Complete settings will disappear. When a step is added the Start and Complete settings will return with their previously saved values.

The start and complete settings are also added when a step is added using the API.

Other settings in the future could include
- a setting to automatically restart the workflow on a given day of the week/month/year e.g. birthday/anniversary notifications and processes that need to repeat periodically.
- a setting to define the message when a workflow is cancelled using the one-click cancel link

**Testing Instructions**
- Acceptance tests add to check the start and complete settings.
- Ensure existing workflows work as expected without adding start or complete steps.
- Add a start step to an existing workflow and test the conditional logic, schedule, pending message and default display fields settings.
- Add a complete step and test the complete message and display fields settings.
- Ensure the start and complete steps stay pinned and that the other steps retain their order.
- Repeat for a new workflow - the steps will be added automatically.
- Remove all the steps - the start and complete steps will be hidden until a step is added again.


includes/steps/class-step.php
```diff
@@ -2230,4 +2230,19 @@ public function purge_assignees() {
 		$this->log_debug( 'Assignees purged. number of rows deleted: ' . $result );
 	}
 
+	public function is_user_assignee() {
```
richardW8k: needs a docblock

class-gravity-flow.php
```diff
@@ -1441,6 +1563,48 @@ public function save_feed_validation_callback( $field, $field_setting ) {
 			return true;
 		}
 
+		public function save_feed_settings( $feed_id, $form_id, $settings ) {
```
richardW8k: needs a docblock

PR #212, opened 2019-02-09
Throttle license checks

stevehenty:
Ensures that only one license check is made for extensions on each request. Also excludes third-party extensions from the admin notices.


stevehenty:
I’m going to take a different approach.

PR #214, opened 2019-02-10
Add filter gravityflow_next_step

Idealien:
Refer to HS#7755

**Issue** 
Provide method for customisation of the next step a workflow will process. Example scenario is  a successfully completed user input step needs to restart the same step in order to trigger different assignee's (based on value changes of current step)

**Closest Alternatives**
- gravityflow_post_process_workflow is too late in workflow cycle to trigger change
- gravityflow_step_complete would allow the GFlow $api->restart_step to modify $entry meta to indicate a successful restart, however it would require use of a custom event (via wp_schedule_single_event) and still not prevent the next ordered workflow step or notification from starting. An additional refresh of the entry during process_workflow does not modify the active step / next step logic enough for this approach either.

**Testing Instructions**
- Create a workflow with multiple steps
- Use this branch and a modified version of the following snippet to have next step be identified as current step.

```
add_filter( 'gravityflow_next_step', 'sh_restart_next_step_workflow', 10,  4 );
function sh_custom_next_step_workflow( $next_step, $current_step, $entry, $steps ) {
	//Modify step ID to match your test scenario
	if ( $next_step != false && $current_step->get_id() == 84 ) {
		//Additional logic would go here to determine if/how $next_step should be modified
		$next_step = $current_step;
	}

	return $next_step;
}
```

### DIFF
--- a/class-gravity-flow.php
+++ b/class-gravity-flow.php
@@ -3543,32 +3543,35 @@ jQuery('#setting-entry-filter-{$name}').gfFilterUI({$filter_settings_json}, {$va
 		/**
 		 * Returns the next step for the supplied entry.
 		 *
-		 * @param Gravity_Flow_Step $step The current step.
+		 * @param Gravity_Flow_Step $step The current step. Defaults to false.
 		 * @param array             $entry The current entry.
 		 * @param array             $form  The current form.
 		 *
 		 * @return bool|Gravity_Flow_Step
 		 */
-		public function get_next_step( $step, $entry, $form ) {
+		public function get_next_step( $step = false, $entry, $form ) {
+			$current_step = $step;
 			$keep_looking = true;
 			$form_id = absint( $form['id'] );
 			$steps = $this->get_steps( $form_id, $entry );
 			while ( $keep_looking && $step ) {
 
 				if ( ! $step instanceof Gravity_Flow_Step ) {
-					return false;
+					$step = false;
 				}
 
 				$next_step_id = $step->get_next_step_id();
 
 				if ( $next_step_id == 'complete' ) {
-					return false;
+					$step = false;
 				}
 
 				if ( $next_step_id == 'next' ) {
 					$step = $this->get_next_step_in_list( $form, $step, $entry, $steps );
 					$keep_looking = false;
-				} else {
+				} 
+				
+				if ( ! in_array( $next_step_id, array( 'complete', 'next' ) ) ) {
 					$step = $this->get_step( $next_step_id, $entry );
 
 					if ( empty( $step ) ) {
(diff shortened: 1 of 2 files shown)


Idealien:
In theory, the specific 'restart' use case could be achieved by adding a refresh of the $entry in process_workflow after $step->end() has been completed. However, the new filter gives more flexibility for wider workflow customizations through code. Another example use case would be reducing multiple outgoing webhook steps to a common endpoint with use of gravityflow_webhook_args to customise webhook payload into a single step that is repeated multiple times.

class-gravity-flow.php
```diff
@@ -3527,32 +3527,35 @@ public function get_current_step( $form, $entry ) {
...
 			$form_id = absint( $form['id'] );
 			$steps = $this->get_steps( $form_id, $entry );
 			while ( $keep_looking && $step ) {
 
 				if ( ! $step instanceof Gravity_Flow_Step ) {
-					return false;
+					$step = false;
 				}
 
 				$next_step_id = $step->get_next_step_id();
```
richardW8k: this will trigger an error if $step is not an instanceof Gravity_Flow_Step so it should probably be moved to an else attached to the above if

class-gravity-flow.php
```diff
@@ -3527,32 +3527,35 @@ public function get_current_step( $form, $entry ) {
...
 
 				if ( $next_step_id == 'complete' ) {
-					return false;
+					$step = false;
```
richardW8k: as we are not returning early here perhaps $keep_looking should be set to false

class-gravity-flow.php
```diff
@@ -3527,32 +3527,38 @@ public function get_current_step( $form, $entry ) {
...
 		 *
-		 * @param Gravity_Flow_Step $step The current step.
+		 * @param Gravity_Flow_Step $step The current step. Defaults to false.
 		 * @param array             $entry The current entry.
 		 * @param array             $form  The current form.
 		 *
 		 * @return bool|Gravity_Flow_Step
 		 */
-		public function get_next_step( $step, $entry, $form ) {
+		public function get_next_step( $step = false, $entry, $form ) {
```
stevehenty: Is it necessary to make `$step` default to `false`?

Idealien: I thought there was an edge case related to expiration or conditional logic that calls the function with $step = false but cannot reproduce what that was. All acceptance tests pass cleanly without the default. 

If there were an edge case from custom code, it would be passing a value for $step into get_next_step regardless. Will remove in next commit.

stevehenty:
This is a rather significant refactor for a minor release. Let's punt to the next major.

PR #218, opened 2019-02-16
Add extension license key constant

stevehenty:
1. Activate the site if the constant is set for core and extensions.
2. Hide the license key settings for extensions if the license key constant is set.
3. Hide the settings tab for extensions if there are no settings on the tab.
4. HIde the license key step if the constant is defined.

Test with the following PRs:
https://github.com/gravityflow/gravityflowchecklists/pull/16
https://github.com/gravityflow/gravityflowflowchart/pull/3
https://github.com/gravityflow/gravityflowpdf/pull/7

### DIFF
--- a/class-gravity-flow.php
+++ b/class-gravity-flow.php
@@ -8146,7 +8146,7 @@ AND m.meta_value='queued'";
 						break;
 				}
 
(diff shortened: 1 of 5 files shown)


richardW8k:
Should the license key setting be removed if the key defined by the constant is invalid?

stevehenty:
Yes, because the constant overrides the settings. However, the admin notice was displaying a link to the settings page so I've suppressed the buttons when the constant is defined.

PR #219, opened 2019-02-17
Reduce DB calls in the get_users_as_choices() method

yoren:
This PR reduces DB calls significantly when calling the `get_users_as_choices()` method in the Gravity_Flow class.

It also removes the 1000 user limit and allows the login name to be used when the display_name is empty.

## Screenshots
With 66 users, when loading the Approval step settings page, the DB calls generated by the `get_users_as_choices()` method, reduced from 81 calls to 10 calls. Here are the screenshots:

With the `develop` branch, the DB calls are like (see bottom corners for the DB calls and execuate time): 
screenshot 2019-02-15 19 15 52
With this branch, the DB calls are like (see bottom corners for the DB calls and execuate time): 
screenshot 2019-02-15 19 16 16


stevehenty:
I removed the 1000 user limit and added the user_login as a fallback for when the display name is not set which we've seen a few times.

stevehenty:
Before you merge this can you add some unit tests, please?

yoren:
@stevehenty The `get_users_as_choices()` method in the field classes is different from the one in Gravity_Flow because it only gets the `$account_choices` but not the other groups.

It makes refactoring this method to Gravity_Flow_Common less meaningful since it doesn't get called elsewhere. So I revert that commit and added a new test for it. Please review again when you have time. Thanks.

tests/unit-tests/test-get-users-as-choices.php
```diff
@@ -0,0 +1,94 @@
+<?php
+
+/**
+ * Testing Gravity_Flow_Common::get_users_as_choices().
```
stevehenty: This will need changing

PR #221, opened 2019-02-20
Fix import message showing when form has no feeds

travislopes:
### Description
This PR resolves an issue where the Gravity Flow import message appears upon importing a form when the form object contains the `gravityflow` feeds array but does not have any feeds.

### Testing Instructions
1. Export two forms, separately, while running Gravity Flow: One with a Gravity Flow feed, one without
1. On `master` branch, import the form with a feed. The import message should appear.
1. On `master` branch, import the form without a feed. The import message should appear.
1. On `fix-import-message` branch, import the form with a feed. The import message should appear.
1. On `fix-import-message` branch, import the form without a feed. The import message should not appear.


stevehenty:
Thanks @travislopes

PR #224, opened 2019-03-15
Fix date-field based conditional routing

Idealien:
See [8206](https://secure.helpscout.net/conversation/788830995/8206?folderId=1776095)

This PR resolves an issue with conditional routing for step assignees involving date fields which were not being correctly evaluated and would return false in many cases they should not.

**Test Instructions**

- Create a form with a date field
- Create a step which uses conditional routing and the date field for condition (all operators supported)
- Verify the expected assignee is assigned to the step.
- While the field format is yyyy-mm-dd through the use of strtotime for evaluation the majority of date formats which users might input into the routing field will also evaluate correctly.

This PR also renames the function hooked to gform_is_value_match (to filter_gform_is_value_match) to allow more field-specific comparisons in the future and corrects the since tag to when it was added to 2.4.1.

### DIFF
--- a/change_log.txt
+++ b/change_log.txt
@@ -1,1 +1,2 @@
 - Fixed the workflows for GP Nested Forms child entries starting before some Gravity Perks extensions had updated the entries.
+- Fixed an issue that prevented conditional routing from correctly matching date field conditions.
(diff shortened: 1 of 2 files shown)


Idealien:
Good suggestions Richard. Latest commits resolve them.

PR #225, opened 2019-03-20
Fix webhook raw merge tags

Idealien:
See [HS#8371](https://secure.helpscout.net/conversation/798625291/8371?folderId=1776095)

When the outgoing webhook step setting Request Body to Raw Request and merge tags were used within the body, any JSON reserved characters ( : " / tab, etc) would not be properly encoded and would not result in the expected data being sent.

Using a late hook to gform_merge_tag_filter this branch/PR ensures that any uses of merge tags have their values json_encoded prior to the webhook being sent.

**Test Instructions**
- Create an outgoing webhook step with Request Body set to Raw and use {:1} or other merge tags in the body
- Submit a form that has JSON reserved characters in the field of merge tag use
- Capture the API outgoing value and [validate the JSON format](https://jsonlint.com/)
- Switch to this branch and re-perform the test. Note that the validation of JSON will pass.
- Consider whether it would funny (or not) as a programmer to name your child JSON instead of Jason

### DIFF
--- a/change_log.txt
+++ b/change_log.txt
@@ -1,1 +1,2 @@
 - Fixed the workflows for GP Nested Forms child entries starting before some Gravity Perks extensions had updated the entries.
+- Fixed the outgoing webhook for steps with raw request body defined to have merge tag values properly escaped.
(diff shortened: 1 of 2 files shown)


richardW8k:
While it works for field merge tags we will run into this issue again when other merge tags are used such as those for post custom fields and user meta. 

I'm not sure if there is a robust way of handling this that handles all merge tags. 

@stevehenty thoughts?

stevehenty:
@richardW8k let's go with this for the moment and document the limitation for the other merge tags while we work on a better solution.

PR #243, opened 2019-04-19
Fix Notification Step with multi-user field assignee(s)

Idealien:
[See HS#8868](https://secure.helpscout.net/conversation/829212605/8868?folderId=1113492)

If a notification step has a multi-user field set as the assignee and entry contains multiple selections the current version will not deliver a notification to the multiple users selected.

**To Test**
* Create a form with one (or more multi-user workflow fields)
* Create a notification step with the multi-user fields as assignee
* Submit an entry and check email delivery (mailhog in local)
* Note that not all expected recipients received the notification
* Switch to fix-multiuser-field-notification branch and repeat notification step
* Receive all expected notifications

### DIFF
--- a/includes/assignees/class-assignee.php
+++ b/includes/assignees/class-assignee.php
@@ -376,41 +376,55 @@ class Gravity_Flow_Assignee {
 		$assignee_type = $this->get_type();
 		$assignee_id   = $this->get_id();
 
-		if ( $assignee_type == 'email' ) {
-			$email                = $assignee_id;
-			$notification['id']   = 'workflow_step_' . $this->step->get_id() . '_email_' . $email;
-			$notification['name'] = $notification['id'];
-			$notification['to']   = $email;
-			$message              = $this->replace_variables( $message );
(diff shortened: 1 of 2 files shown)


includes/assignees/class-assignee.php
```diff
@@ -376,41 +376,55 @@ public function send_notification( $notification ) {
...
+
+			case 'role':
+				$users = get_users( array( 'role' => $assignee_id ) );
+				break;
 
-			return;
-		}
+			case 'assignee_multi_user_field':
+				$entry = $this->step->get_entry();
+				if ( isset( $entry[ $assignee_id ] ) ) {
```
richardW8k: `! empty` would be better here. using`isset` is resulting in the notification being sent to all users when the entry value is empty.

yoren:
Hey @Idealien can you get this reviewed again and merged into `master`? I'll have to apply to

yoren:
sorry I accidentally closed it for a second.

PR #249, opened 2019-05-04
Add support form multiple forms in the status shortcode

stevehenty:
This PR adds support for multiple form IDs to be used in the status shortcode. e.g. [gravityflow page="status" form="1,2"]

It also adds support enqueuing scripts when the Gravity Flow blocks are added.

**Testing instructions**
Check for regressions with no forms and with one form especially edge cases that use the filters. Then check that the status shortcode works as expected with multiple form IDs in the `form` attribute.

### DIFF
--- a/class-gravity-flow.php
+++ b/class-gravity-flow.php
@@ -811,7 +811,7 @@ PRIMARY KEY  (id)
 
 			$shortcode_found = false;
 			foreach ( $wp_query->posts as $post ) {
-				if ( stripos( $post->post_content, '[gravityflow' ) !== false ) {
+				if ( stripos( $post->post_content, '[gravityflow' ) !== false || stripos( $post->post_content, '<!-- wp:gravityflow/inbox' ) !== false ) {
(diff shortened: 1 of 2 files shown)


class-gravity-flow.php
```diff
@@ -6201,8 +6205,12 @@ public function get_shortcode_status_page( $a ) {
 			);
 
 			if ( isset( $a['form'] ) ) {
+			    $form_id = $a['form'] ? explode( ',', $a['form'] ) : '';
+			    if ( is_array( $form_id ) && count( $form_id ) === 1 ) {
+				    $form_id = $form_id[0];
+                }
 				$args['constraint_filters'] = array(
```
Idealien: Should this convert or validate that array elements are integers and/or actual form IDs?
Test data "apple, , 29" led to constraints of 0,0,29 making it to the GFAPI::get_entries in status prepare_items function.

stevehenty: Yes, we can do this in status prepare_items

PR #250, opened 2019-05-06
Set default `number` to 2000 when get users in the Assignee field/setting

yoren:
This PR fixes an issue that after PR #219, sites with a large number of users cannot load (or load very slowly) step settings or assignee fields.

We now set the default number to `2000` and add a notice to settings description when we don't display a full list of users. We also properly handle a situation that prevents assignees from being removed after the user list gets modified.

## Screenshots
Screenshot 2019-05-12 11 44 45

Screenshot 2019-05-12 11 45 41

## Testing instructions
1. You can use FakerPress to generate more than 2,000 users. Or just use `gravityflow_get_users_args` to set a default `number`, so you can see the new description in your assignee and routing assignees fields.
2. Set some steps with assignees, pick some users from the bottom of the user list. Now use the `gravityflow_get_users_args` filter to change the `number` to `5` (to make sure the assignees previously picked won't display in the user list). Now reload the step settings, we expect to see them still show up in the user list.
3. Be sure you test both assignees and routing assignees. Also test the Notification step and different notification tabs in Approval/User Input steps etc..

## Note
Gravity Flow PDF extension loads feed settings as the step settings so we need to manually add the new description to it. I will create another PR for it.


class-gravity-flow.php
```diff
@@ -8655,5 +8701,22 @@ public function action_admin_notices() {
 				GFCommon::display_dismissible_message( $notices );
 			}
 		}
+
+		/**
+		 * Helper function to return gravityflow_get_users_args filter value.
+		 *
+		 * @since 2.6
```
stevehenty: I'm going to release this in the next minor so the since tags can be 2.5.3.

class-gravity-flow.php
```diff
@@ -8655,5 +8701,22 @@ public function action_admin_notices() {
...
+		 *
+		 * @return array
+		 */
+		public function get_users_args() {
```
stevehenty: Can we put this in `Gravity_Flow_Common`?

includes/steps/class-common-step-settings.php
```diff
@@ -400,12 +427,21 @@ public function get_setting_assignees() {
...
 	public function get_setting_assignee_routing() {
-		return array(
+		$setting = array(
 			'name'    => 'routing',
 			'tooltip' => __( 'Build assignee routing rules by adding conditions. Users and roles fields will appear in the first drop-down field. If the form contains any assignee fields they will also appear here. Select the assignee and define the condition for that assignee. Add as many routing rules as you need.', 'gravityflow' ),
 			'label'   => __( 'Routing', 'gravityflow' ),
 			'type'    => 'routing',
 		);
+
+		$total_accounts = self::$_total_accounts;
```
stevehenty: I think `self::$_total_accounts` can use `$this`?

includes/steps/class-common-step-settings.php
```diff
@@ -390,6 +407,16 @@ public function get_setting_assignees() {
 			'type'     => 'select',
 			'choices'  => $this->_account_choices,
 		);
+
+		$total_accounts = self::$_total_accounts;
+		$args           = gravity_flow()->get_users_args();
+		$number         = ( isset( $args['number'] ) && $args['number'] > 0 ) ? $args['number'] : 2000;
+		if ( $total_accounts['total_users'] > $number ) {
+			/* translators: 1: Warning icon 2: Number of users displayed 3: Open link tag 4: Close link tag */
+			$setting['description'] = sprintf( esc_html__( '%1$s This list of users contains only the first %2$s users in your site. %3$sLearn how to remove this limit%4$s. ', 'gravityflow' ), '<i class="dashicons dashicons-warning" style="color:red;"></i> ', $number, '<a href="https://docs.gravityflow.io/article/54-gravityflowgetusersargs" target="_blank">', '</a>' );
```
stevehenty: I've adjusted the message and added a note for the translators. Can you apply it to the other instances, please?

includes/steps/class-common-step-settings.php
yoren: `$_total_accounts` was a static var but I ended up to move it to Gravity_Flow_Common as a static function too.

yoren:
@stevehenty @richardW8k This PR needs some more work because I noticed the changes haven't covered the notification tabs (approval/rejection emails etc.). I'm now working on it and will ping you again when it's ready.

yoren:
Hi @stevehenty @richardW8k this PR is ready for your review now. Thanks.

stevehenty:
Since the refactor something's gone a bit wrong. An approval step is assigned to a user which is not included in the list due to the filter however the user isn't showing in the list. Before the refactor the assignee was loading despite not being in the list.

yoren:
@stevehenty I think it's because I added a check for if the notification emails are enabled. I remove that and it works for me now.

stevehenty:
@richardW8k Which notification is it missing for? I'm seeing the warnings in the approval emails and the notification step workflow emails.

stevehenty:
@richardW8k got it - yeah, it's not appearing on the conditional routing for the notification step.

yoren:
@stevehenty @richardW8k I've found the cause of the issue and it has been fixed now

PR #259, opened 2019-06-08
Fix order summary on user input steps without product fields

Idealien:
Refer to [HS#9207](https://secure.helpscout.net/conversation/851721997/9207?folderId=1776095#)

User Input steps cannot currently display the order summary without having a product field edit/display. The step settings already include option whether to display Order Summary.
image

This PR checks whether that is set as an additional reason to display the Order Summary.

See HS#9207 for test form/instructions.

### DIFF
--- a/change_log.txt
+++ b/change_log.txt
@@ -1,1 +1,1 @@
--Fixed scenario where user input steps would not show the product order summary table when step setting was selected
+- Fixed an issue with the user input step where the product order summary table does is not displayed unless a product field is included in the display fields.
--- a/change_log.txt
+++ b/change_log.txt
@@ -1,0 +1,1 @@
+-Fixed scenario where user input steps would not show the product order summary table when step setting was selected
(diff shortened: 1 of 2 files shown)


change_log.txt
```diff
@@ -0,0 +1 @@
+-Fixed scenario where user input steps would not show the product order summary table when step setting was selected
```
stevehenty: Fixed an issue with the user input steps where the product order summary table does is not displayed unless a product field is included in the display fields.

includes/pages/class-entry-detail.php
```diff
@@ -770,7 +770,7 @@ public static function entry_editor( $form, $entry, $current_step, $display_empt
 				</td>
 			</tr>
 			<?php
-			if ( $entry_editor->has_product_fields ) {
+			if ( $entry_editor->has_product_fields || $current_step->display_order_summary ) {
```
stevehenty: Can we move both these checks into `maybe_show_products_summary()`?

PR #261, opened 2019-06-18
Fix delayed Zapier steps send duplicate content from the first entry in queue

yoren:
re:[HS#9536](https://secure.helpscout.net/conversation/868343234/9536?folderId=1113497)

This PR fixes an issue that delayed Zapier steps send duplicate content from the first entry in queue, when there are multiple scheduled entries.

## Testing instructions
Please reproduce the error as the ticket described, and it will be fixed with this branch.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM php:7.2-cli
+FROM php:7.2-cli-stretch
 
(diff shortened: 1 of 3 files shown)


yoren:
hey @Idealien - can you review this PR when you got a minute? Thanks!

PR #265, opened 2019-07-20
Add HubSpot step for the Gravity Forms HubSpot add-on

yoren:
This PR adds a new HubSpot step for the Gravity Forms HubSpot add-on.

## Testing instructions
1. Create a form with a HubSpot feed.
2. Add a new HubSpot step to the workflow, and add an Approval step before it.
3. After submitting the form, make sure the email only get subscribed after approval.


yoren:
@Idealien Yeah because there are two different HubSpot add-on for Gravity Forms. We've supported the older one, and RG just released the official HubSpot add-on after we partner with HubSpot.

PR #270, opened 2019-08-20
Support step specific complex conditional logic

Idealien:
See [HS#10486](https://secure.helpscout.net/conversation/932770262/10509/)

This PR adds the filter gravityflow_step_is_condition_met and customizes the logic within the is_condition_met function. Prior to this PR:
- You were limited to a single ANY (or) / ALL (and) set of conditions for enabling condition
- If you wanted to evaluate step conditions through code, you had to set a known false conditional setting in order to evaluate through [gform_is_value_match](https://docs.gravityforms.com/gform_is_value_match/).

**Test Instructions**
-Switch to the PR branch in your local environment
-Create a form with 2 workflow steps - 1 with conditional setting and 1 without
-Add the a variation of the following snippet to your functions.php file. This one assumes some step, form and field IDs but will value check a different entry field to determine if the current potential step should be executed.
```
add_filter( 'gravityflow_step_is_condition_met', 'gform_109_7_1_conditional_logic_match_14_9'), 10, 5);

function gform_109_7_conditional_logic_match_14_9( $do_action, $logic, $form, $entry, $step ) {
		if ( false == $step ) {
			return false;
		}

		$step_id = $step->get_id();
		if ( false == in_array( $step_id, array( '109', '111' ) ) ) {
			return false;
		}
		if( !empty( $entry['7'] ) ) {
			$search_criteria['field_filters'][] = array( 'key' => '1', 'value' => $entry['7'] );
			$paging      = array( 'offset' => 0, 'page_size' => 1 );
			$total_count = 0;
			$apps        = GFAPI::get_entries( 14, $search_criteria, array(), $paging, $total_count );
			if( ! empty( $apps ) ) {
				foreach( $apps as $app ) {
					if( isset( $app['19'] ) && $app['19'] == 'Y' ) {
						$is_match = true;
					} else {
						$is_match = false;
					}
				}
			}
		}
		return $is_match;
	}
```

### DIFF
--- a/change_log.txt
+++ b/change_log.txt
@@ -4,3 +4,4 @@
 - Fixed number field display on user input steps when the field contained 0 value.
 - Fixed delayed Zapier steps send duplicate content from the first entry in queue.
 - Added Merge Tag selector to Outgoing Webhook step settings for raw request body.
+- Added the filter gravityflow_step_conditional_logic_match to enable complex conditional logic for step execution.
--- a/change_log.txt
+++ b/change_log.txt
@@ -4,4 +4,4 @@
 - Fixed number field display on user input steps when the field contained 0 value.
 - Fixed delayed Zapier steps send duplicate content from the first entry in queue.
 - Added Merge Tag selector to Outgoing Webhook step settings for raw request body.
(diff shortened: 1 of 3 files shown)


Idealien:
Was started from master branch. Will re-submit with a develop-based branch.

PR #278, opened 2019-09-22
Extend settings + UI to always show CC field

Idealien:
See [HS #10520](https://secure.helpscout.net/conversation/933782946/10520?folderId=1113497)

This PR adds the option for CC: to be included for all notification types of Gravity Flow (assignee, approval, rejection, complete). It does not currently evaluate whether [gform_notification_enable_cc](https://docs.gravityforms.com/gform_notification_enable_cc/) filter has been set or not. Perhaps worth discussion whether it should, or a Gravity Flow filter that includes step-specific parameters.

**Testing Instructions**

- Enable the dev-notification-cc branch
- Test that all notification types include an option for CC: distribution
- Confirm that notifications deliver with CC: users/emails specified


Idealien:
fix-10520-enable-cc branch for gravityflowpdf is already created.
It would be good to review/release all of the add-ons with the settings updates at same time.

stevehenty:
The CC setting needs to be hidden when the email is not enabled.

Screen Shot on 2019-09-29 at 15-33-43

stevehenty:
Sorry, false alarm, my browser was caching the JS file.

PR #282, opened 2019-10-11
Support API send_to_step scenarios when proposed step conditions not met

Idealien:
See [HS#10796](https://secure.helpscout.net/conversation/948695322/10796?folderId=1776095) which originally relates to step options available in the admin actions which can be customized by [gravityflow_admin_actions_workflow_detail](https://docs.gravityflow.io/article/108-gravityflowadminactionsworkflowdetail) filter which relies on orchestration API send_to_step functions. The current behaviour is to send the entry/workflow to the proposed new step regardless of whether the new step will pass its' is_condition_met check.

This PR maintains backwards compatibility yet offers users the ability to customize if and to what step a workflow should be sent to in the case where the is_condition_met check for the proposed new step is not met. It will loop through every subsequent step in workflow to find a valid one if it exists, or leave the workflow at the current step if none do.

**Testing Instructions**
- Download and import this test form to your local environment [gravityforms-export-2019-10-11.zip](https://github.com/gravityflow/gravityflow/files/3715395/gravityforms-export-2019-10-11.zip)
- Switch to the PR branch
- Preview/submit a version of the form
- Confirm that the workflow goes through/skips expected steps
- Use the admin actions menu on entry detail page to send to various steps. Confirm that all go to the chosen step regardless of conditions being met or not.
- Write a function that uses [send to step API](https://docs.gravityflow.io/article/90-workflow-orchestration-api#send_to_step) and confirm the workflow is sent to desired step regardless of conditions.
- Add the following code to your environment.
```
add_filter( 'gravityflow_send_to_step_condition_met_required', 'jo_send_to_step_condition_not_required', 10, 5 );
function jo_send_to_step_condition_not_required( $conditions_met_required, $new_step, $current_step, $entry, $form ) {
	$conditions_met_required = true;
	return $conditions_met_required;
}
```
- Repeat previous test scenarios and note that any where the condition is not met for a proposed step, the workflow timeline will include a ""Send condition not met to send to step: [Chosen step title]" message. The workflow will either be sent to the next step after chosen one or leave the workflow unchanged if any subsequent step also cannot meet its conditions.
- Add the following code to your environment and customize what step is returned.
```
add_filter( 'gravityflow_send_to_step_next_step', 'jo_what_step_next', 10, 5 );
function jo_what_step_next( $next_step, $new_step, $current_step, $entry, $form ) {
	$api = new Gravity_Flow_API( $form['id'] );
	/* @var Gravity_Flow_Step[] $steps */
	$steps = $api->get_steps();

	//Decide which step to go to instead of the one which failed it's conditions (new step)
	return $next_step;
}
```
- Repeat previous test scenarios and note that when condition not met for a chosen step that your custom logic executes as expected.

**Code Review Questions**
1. Do you agree that the naming convention of filters/variables added makes sense?
There is a non-trivial amount of overlap of what current, new, next steps mean in context of sending a workflow to a different step.
1. This PR is maintaining backwards compatibility by gravityflow_send_to_step_condition_met_required being default to false. Do you agree that is the expected behaviour?
1. Should gravityflow_send_to_step_next_step only occur when gravityflow_send_to_step_condition_met_required has been customized to return true (as PR now has it). An iteration where it occurred in both led to the confusing potential filter gravityflow_send_to_step_step or gravityflow_send_to_step_next_step which just seemed wrong.


includes/class-api.php
```diff
@@ -225,6 +225,73 @@ public function get_status( $entry ) {
 	public function send_to_step( $entry, $step_id ) {
 		$current_step = $this->get_current_step( $entry );
 		if ( $current_step ) {
+
+			$new_step = $this->get_step( $step_id, $entry );
+			$form     = $new_step->get_form();
```
richardW8k: `$new_step` could be `false`, so that needs checking before attempting to get the form.

includes/class-api.php
```diff
@@ -225,6 +225,73 @@ public function get_status( $entry ) {
...
+				*
+				* @param bool                   $conditions_met_required  Whether to pass a workflow to a step that has failed its required conditions.
+				* @param Gravity_Flow_Step      $new_step                 The proposed new step that failed its step conditions.
+				* @param Gravity_Flow_Step      $current_step             The current step.
+				* @param array                  $entry                    The current entry.
+				* @param array                  $form                     The current form.
+				*
+				* @return bool
+				*/
+				$conditions_met_required = apply_filters( 'gravityflow_send_to_step_condition_met_required', false, $new_step, $current_step, $entry, $form );
```
richardW8k: Any reason the filter can't be located before the if so it can prevent the condition check being performed when not required?

Idealien: Good feedback, thx. I updated placement in latest commit, along with added validation on its' return value. Also, the filter name "gravityflow_send_to_step_next_step" doesn't reflect that it is tied to the failure of $next_step to pass conditions. Renamed to gravityflow_send_to_step_condition_not_met but would appreciate feedback if you think further clarification required.

PR #285, opened 2019-10-26
Fix PHP fatal error thrown when processing expired entries with steps that no longer exist

yoren:
re:[HS#11058](https://secure.helpscout.net/conversation/974045212/11058?folderId=1113492)

This PR fixes an issue where PHP fatal error is thrown when processing expired entries with steps that no longer exist.

## Testing instructions
This is a fix that we add based on the PHP fatal error message the ticket provided. We don't seem to find an easy way to reproduce the error at our end.

### DIFF
--- a/change_log.txt
+++ b/change_log.txt
@@ -1,2 +1,3 @@
 - Fixed incorrect feedback when user input step confirmation message content is left empty and the step is completed.
 - Fixed an issue where PHP fatal error is thrown in the filter_gravityview_common_get_entry_check_entry_display() method.
+- Fixed an issue where PHP fatal error is thrown when processing expired entries with steps that no longer exist.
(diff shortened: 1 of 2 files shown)


yoren:
hey @Idealien 

> Did you try to test through de-activated plugin for the step? Either prior to the maybe_process_expiration_and_reminders execution or with a larger data set that de-activation could happen during execution?

No I didn't try this method. I'd say with a larger data set the error should be able to be reproduced ("prior to the maybe_process_expiration_and_reminders execution" wouldn't work since the step won't be checked at all in the first place). But that'll be a very edgy case IMO. Our discussion here should be a good reference if relevant issues are brought up in the future.

Will merge this PR.

PR #295, opened 2019-12-05
Prevent signature warning with GravityView 2.5

soulseekah:
`GravityView_Field::field_options` has a 5th parameter called `$form_id` since GravityView 2.5

https://github.com/gravityview/GravityView/commit/5e6af4d6ce087e0d199f520159d0814172500b3b


Idealien:
Thanks for the PR @soulseekah!
I've added attribution into the change_log.txt which we aim for with any bug/fix/enhance for release.

PR #303, opened 2020-02-20
Add shortcode processing to outgoing webhook raw body requests

Idealien:
## Description
Enhance the outgoing webhook to allow the use of shortcodes such as [conditionals](https://docs.gravityforms.com/conditional-shortcode/) within the raw body structure. In order for a similar change to be applied through the field mapping setting, Gravity Forms issue #807 will have to be addressed regarding parsing " within the generic_map settings field type.

## Testing instructions

- Create a form/workflow involving outgoing webhook
- Use the [conditional shortcode](https://docs.gravityforms.com/conditional-shortcode/) for content such as
```
[gravityforms action="conditional" merge_tag="{Number:1}" condition="greater_than" value="10"]
"dynamic-key": "would only be displayed when field 1 > 10",
[/gravityforms]
```
- Note that master branch does not evaluate the shortcode although the merge tag within it is.
- Change to enhance-webhook-conditionals branch and re-run the webhook step
- Confirm that the conditional value is shown/hidden as appropriate.
- Repeat with other shortcode types

## Documentation Changes?
Update [Outgoing Webhook Step Notes](https://docs.gravityflow.io/article/114-the-outgoing-webhook-step) to identify that shortcodes are supported.

## Checklist:
- [x] I've tested the code.

### DIFF
--- a/includes/steps/class-step-webhook.php
+++ b/includes/steps/class-step-webhook.php
@@ -614,7 +614,7 @@ class Gravity_Flow_Step_Webhook extends Gravity_Flow_Step {
 			$body = $this->raw_body;
 			add_filter( 'gform_merge_tag_filter', array( $this, 'filter_gform_merge_tag_webhook_raw_encode' ), 40, 5 );
 			$body = GFCommon::replace_variables( $body, $this->get_form(), $entry, false, false, false, 'text' );
-			$body = GFCommon::gform_do_shortcode( $body );
+			$body = do_shortcode( $body );
 			remove_filter( 'gform_merge_tag_filter', array( $this, 'filter_gform_merge_tag_webhook_raw_encode' ), 40 );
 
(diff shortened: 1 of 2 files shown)


includes/steps/class-step-webhook.php
```diff
@@ -614,6 +614,7 @@ function send_webhook() {
 			$body = $this->raw_body;
 			add_filter( 'gform_merge_tag_filter', array( $this, 'filter_gform_merge_tag_webhook_raw_encode' ), 40, 5 );
 			$body = GFCommon::replace_variables( $body, $this->get_form(), $entry, false, false, false, 'text' );
+			$body = GFCommon::gform_do_shortcode( $body );
```
richardW8k: I would use the standard `do_shortcode` here as this context doesn't need the extra overhead of checking for embedded forms and including the scripts and styles.

PR #310, opened 2020-03-31
Fix Duplicate HubSpot Step

saifsultanc:
## Description
This fixes the issue of duplicate HubSpot steps. The issue of having steps displayed for both Gravity Forms Hub Spot Add-on and the third party Hub Spot Add-on is resolved.

## Testing instructions
- Install the Offical Gravity Forms Hub Spot Add-on.
- Install the third-party Gravity Forms Hub Spot Add-on: https://wordpress.org/plugins/gravityforms-hubspot/
- Activate only the Official Gravity Forms Hub Spot Add-on and check workflow step only consists of this one and not the other.
- Activate only the third-party Gravity Forms Hub Spot Add-on and check workflow step only consists of this one and not the other.
- Activate both Add-ons and check workflow step contains both.

N/A
 
## Screenshots
N/A

## Documentation Changes?
N/A

## Checklist:
- [x] I've tested the code.

### DIFF
--- a/change_log.txt
+++ b/change_log.txt
@@ -19,4 +19,4 @@
 - Fixed an issue where the report data for assignees by month is incorrect.
 - Fixed an issue where the Discussion Field view more/less effect would not operate on the Form > Entries screen (both view and edit).
 - Fixed an issue where the scripts and styles are not enqueued when using the shortcode or block in a reusable block.
-- Fixed an issue of duplicacy on Workflow steps with Gravity Forms Hub Spot Add-on and third party Hub Spot Add-on.
+- Fixed an issue in the step settings page where duplicate Workflow step icons appear for the Gravity Forms HubSpot Add-on and third-party HubSpot Add-on. IMPORTANT: check that your HubSpot workflow steps are correct after updating to this version.
--- a/change_log.txt
+++ b/change_log.txt
@@ -19,3 +19,4 @@
 - Fixed an issue where the report data for assignees by month is incorrect.
 - Fixed an issue where the Discussion Field view more/less effect would not operate on the Form > Entries screen (both view and edit).
 - Fixed an issue where the scripts and styles are not enqueued when using the shortcode or block in a reusable block.
+- Fixed an issue of duplicacy on Workflow steps with Gravity Forms Hub Spot Add-on and third party Hub Spot Add-on.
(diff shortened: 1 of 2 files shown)


richardW8k:
I wonder if anyone has configured the third-party step to process feeds from the official add-on? In which case this change will prevent the step being processed.

saifsultanc:
Hi @richardW8k 

Thanks for having a look at this. Please correct me if I understand this correctly, or if I am missing any detail...

The condition `if ( ! class_exists( '\BigSea\GFHubSpot\GF_HubSpot' ) ) ` is there to ensure that when the third-party add-on isn't there, the class name 'GF_HubSpot' is used which will be for the official add-on. Thus, when only the official add-on is installed, this condition created a duplicate instance for the 'GF_HubSpot'. The other one is coming in from the steps class of the official add-on (Gravity_Flow_Step_Feed_GFHubSpot). On removing this condition, it ensured that the third party ran for itself only, and likewise for the official one. Therefore, no duplicates on the Workflow steps, which was the purpose of the Trello task.

Also, if you can mention any other input, it would be helpful.

Thanks

richardW8k:
@saifsultanc the `hubspot` step was added in 2016. Back then the third-party add-on used the class name `GF_HubSpot`. In 2017 the third-party add-on changed to using namespaces so to support both versions of the add-on the `get_feed_add_on_class_name` method was added to determine which class name to use. In 2019 Gravity Forms released it's HubSpot add-on, which uses the `GF_HubSpot` class name, so the `gfhubspot` type step was added.

As the `hubspot` step uses  the `GF_HubSpot` class name when the third-party add-on is not active and the official add-on is active there is a chance some users could have used it instead of the `gfhubspot` step.

I've discussed it with Steve; for now we are going to handle it via a note in the change log and docs.

change_log.txt
```diff
@@ -19,3 +19,4 @@
 - Fixed an issue where the report data for assignees by month is incorrect.
 - Fixed an issue where the Discussion Field view more/less effect would not operate on the Form > Entries screen (both view and edit).
 - Fixed an issue where the scripts and styles are not enqueued when using the shortcode or block in a reusable block.
+- Fixed an issue of duplicacy on Workflow steps with Gravity Forms Hub Spot Add-on and third party Hub Spot Add-on.
```
stevehenty, line 22:
```suggestion
- Fixed an issue of duplicate Workflow steps for the Gravity Forms HubSpot Add-on and third-party HubSpot Add-on. IMPORTANT: check that your HubSpot workflow steps are correct after updating to this version.
```

saifsultanc:
Thanks @richardW8k  and @stevehenty.

PR #313, opened 2020-04-18
Add support for Zapier Add-On v4.0

richardW8k:
## Description
This updates the Zapier step to support processing feeds for the famework version of the add-on. As the feed IDs change during the migration the steps will also be updated at the end of the migration to ensure selected feeds will continue to be processed by the step.

## Testing instructions
- Use this branch and Zapier master branch (v3.2.1)
- Create a form with a Zapier feed
- Configure a workflow with an Approval step and a Zapier step
- Test the form to confirm the zap is only processed after the Approval step
- Update to the Zapier 4.0 branch (v4.0-beta-1.1)
- Check the Zapier step to confirm the feed is selected
- Test the form again to confirm the zap is only processed after the Approval step
- Check a form which does not have a Zapier feed to confirm Zapier is not an available type when creating a new step

N/A
 
## Screenshots
N/A

## Documentation Changes?
N/A

## Checklist:
- [x] I've tested the code.

### DIFF
--- a/includes/steps/class-step-feed-zapier.php
+++ b/includes/steps/class-step-feed-zapier.php
@@ -216,22 +216,23 @@ class Gravity_Flow_Step_Feed_Zapier extends Gravity_Flow_Step_Feed_Add_On {
 				continue;
 			}
 
-			$step_dirty = false;
+			$to_migrate = array();
 			$step_meta  = $step->get_feed_meta();
 
 			foreach ( $migrated_feeds as $feed ) {
 				$legacy_id = rgars( $feed, 'meta/legacy_id' );
 				if ( $legacy_id && isset( $step_meta[ 'feed_' . $legacy_id ] ) ) {
-					$step_dirty = true;
-					$enabled    = $step_meta[ 'feed_' . $legacy_id ] == '1';
+					$to_migrate[ $legacy_id ] = $step_meta[ 'feed_' . $legacy_id ] == '1' ? $feed['id'] : false;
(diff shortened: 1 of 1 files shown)


includes/steps/class-step-feed-zapier.php
```diff
@@ -60,74 +71,175 @@ public function get_icon_url() {
...
+				$legacy_id = rgars( $feed, 'meta/legacy_id' );
+				if ( $legacy_id && isset( $step_meta[ 'feed_' . $legacy_id ] ) ) {
+					$step_dirty = true;
+					$enabled    = $step_meta[ 'feed_' . $legacy_id ] == '1';
+					unset( $step_meta[ 'feed_' . $legacy_id ] );
+					if ( $enabled ) {
+						$step_meta[ 'feed_' . $feed['id'] ] = '1';
+					}
+				}
+			}
```
yoren, line 232: hey @richardW8k - I've tested and it worked as expected. My only concern is how we set up the feed ID replacement here. Could it be a possibility that the following example:

Legacy feed ID 5 => Migrated feed ID 7
Legacy feed ID 7 => Migrated feed ID 9

So the step meta used to have feed_5 and feed_7, but after the loop it becomes having feed_9 only.

I didn't really test so I could be wrong. Let me know what you think. Thanks!

richardW8k: @yoren Good catch. The latest commit handles that scenario.

PR #314, opened 2020-04-22
Show empty editable fields in {workflow_fields} for the User Input Step

saifsultanc:
## Description
Added a merge tag modifier to allow the empty fields to be displayed in the {workflow_fields} merge tag. e.g. {workflow_fields: show_empty=true}.
This helps when working with the User Input step, for instance. Only the empty and editable fields are displayed.

Reference:
https://trello.com/c/zTPq5WvC/518-show-empty-fields-in-workflowfields
https://gravityflow.productboard.com/insights/all-notes/notes/7480020
https://secure.helpscout.net/conversation/1016383996/11740/

## Testing instructions

1. Create a Form with two fields 'Field-1' and 'Field-2'.
2. Create a Notification step (Workflow step# 1). Make the current user as Assignee, and on the email message body include the merge tag: `{workflow_fields: show_empty=true}`
3. Create a User Input step (Workflow step# 2) with 'Field-2' as editable. Make the current user as Assignee, and on the email message body include the merge tag: `{workflow_fields: show_empty=true}`
4. Run the Form and enter some value for 'Field-1' leaving 'Field-2' empty.
5. Check emails for the user. The email of Workflow step# 1 will not have any data for it.
6. The email of Workflow step# 2 will have the 'Field-2' listed as the empty editable entry.

N/A
 
## Screenshots
N/A

## Documentation Changes?
Update documentation about the merge tag modifier here: https://docs.gravityflow.io/article/106-merge-tags

## Checklist:
- [x] I've tested the code.

### DIFF
--- a/includes/merge-tags/class-merge-tag-workflow-fields.php
+++ b/includes/merge-tags/class-merge-tag-workflow-fields.php
@@ -63,6 +63,7 @@ class Gravity_Flow_Merge_Tag_Workflow_Fields extends Gravity_Flow_Merge_Tag {
 				$modifiers = rgar( $match, 2 );
 
 				$a = $this->get_attributes( $modifiers, array(
+					'show_empty'	=> false,	// Output empty fields (for User Input step).
 					'empty'    => false, // Output empty fields.
 					'value'    => false, // Output choice values.
 					'admin'    => false, // Output admin labels.
@@ -109,8 +110,20 @@ class Gravity_Flow_Merge_Tag_Workflow_Fields extends Gravity_Flow_Merge_Tag {
 	 */
 	public function merge_tag_filter( $value, $merge_tag, $modifiers, $field ) {
 		$modifiers_array        = $field->get_modifiers();
+		$display_show_empty			= in_array( 'show_empty', $modifiers_array ) && Gravity_Flow_Common::is_editable_field( $field, $this->step );
 		$display_editable_field = in_array( 'editable', $modifiers_array ) && Gravity_Flow_Common::is_editable_field( $field, $this->step );
 		$display_display_field  = in_array( 'display', $modifiers_array ) && Gravity_Flow_Common::is_display_field( $field, $this->step, $this->form, $this->entry );
+
+		if ( in_array( 'show_empty', $modifiers_array ) ) {
+			// Only display editable fields with 'show_empty' for a User Input step.
+			if ( $display_show_empty ) {
+				$value = ' ';
+				return $value;
+			}
+			else {
+				return false;
+			}
+		}
 
 		if ( ! $display_editable_field && ! $display_display_field ) {
 			// Removing non-editable and non-display field from merge tag output.
--- a/includes/merge-tags/class-merge-tag-workflow-fields.php
+++ b/includes/merge-tags/class-merge-tag-workflow-fields.php
@@ -71,7 +71,7 @@ class Gravity_Flow_Merge_Tag_Workflow_Fields extends Gravity_Flow_Merge_Tag {
 					'display'  => true, // Output the steps display fields.
 				) );
 
-				$replacement = GFCommon::get_submitted_fields( $this->form, $entry, $a['empty'], ! $a['value'], $this->format, $a['admin'], $this->name, $this->get_options_string( $a ) );
+				$replacement = GFCommon::get_submitted_fields( $this->form, $entry, $a['empty'], ! $a['value'], $this->format, $a['admin'], $this->name, $modifiers );
 				$text        = str_replace( $full_tag, $replacement, $text );
 			}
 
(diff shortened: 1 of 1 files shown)


includes/merge-tags/class-merge-tag-workflow-fields.php
```diff
@@ -109,9 +110,21 @@ public function get_options_string( $attributes ) {
 	 */
 	public function merge_tag_filter( $value, $merge_tag, $modifiers, $field ) {
 		$modifiers_array        = $field->get_modifiers();
+		$display_show_empty			= in_array( 'show_empty', $modifiers_array ) && Gravity_Flow_Common::is_editable_field( $field, $this->step );
```
Idealien, line 113: When using this on notification step, the empty fields (or any of the merge tag) do not show up. Did you mean to use is_display_field instead of is_editable_field ?

includes/merge-tags/class-merge-tag-workflow-fields.php
```diff
@@ -109,9 +110,21 @@ public function get_options_string( $attributes ) {
...
 		$display_editable_field = in_array( 'editable', $modifiers_array ) && Gravity_Flow_Common::is_editable_field( $field, $this->step );
 		$display_display_field  = in_array( 'display', $modifiers_array ) && Gravity_Flow_Common::is_display_field( $field, $this->step, $this->form, $this->entry );
 
+		if ( in_array( 'show_empty', $modifiers_array ) ) {
+			// Only display editable fields with 'show_empty' for a User Input step.
+			if ( $display_show_empty ) {
+				$value = ' ';
```
Idealien, line 120: Compare to other merge tags such as class-merge-tag-current-step.php - the return value should be blank when merge tag will not be producing any output.

includes/merge-tags/class-merge-tag-workflow-fields.php
saifsultanc: Hi @Idealien, From what I understood, the intent of the 'show_empty' was to show **_editable_** fields albeit empty. Following the 'insights' on PB linked to [here](https://secure.helpscout.net/conversation/1016383996/11740/#thread-2874651566).

Comparing the two steps 'Notification' and 'User Input', there are no editable fields for a notification, hence the blank output. 

Please suggest if I am missing something here.

Thanks, Jamie.

Idealien: Understood. Both is_display_field and is_editable_field apply largely to the user input step / entry detail screen and within this merge tag already have their own identifiers. I'm not sure that naming the new parameter show_empty makes sense when :empty can already be defined.

Perhaps is a new modifier not required, but just the logic in the merge tag changed to address [using multiple modifiers](https://docs.gravityforms.com/fields-merge-tag/)? So {workflow_fields:editable,empty} being the immediate need and ensuring {workflow_fields:display,empty} does as well.

includes/merge-tags/class-merge-tag-workflow-fields.php
```diff
@@ -63,6 +63,7 @@ public function replace( $text ) {
 				$modifiers = rgar( $match, 2 );
 
 				$a = $this->get_attributes( $modifiers, array(
+					'show_empty'	=> false,	// Output empty fields (for User Input step).
```
Idealien, line 66: Using spaces instead of tabs follows the WPCS standards for multidimensional array notation.

includes/merge-tags/class-merge-tag-workflow-fields.php
saifsultanc: Hi @Idealien 

Made some updates on this. Do you think I am in the right direction?

There's 1 thing I don't know whether I'm missing something in understanding it or an inadvertent issue on our codebase at line 73 [here](https://github.com/gravityflow/gravityflow/blob/a533d13531ea2cdfd910d2e81e4fa4edc2bf677e/includes/merge-tags/class-merge-tag-workflow-fields.php#L73).

Why do we send the `$a`, which is hard-coded above on the lines 65-71?
Why don't we pass the `$modifiers` of line 63 (which would be coming as per the workflow step settings done by the user)?

Any other things I can do on the code to improve the logic? Or have I messed up the logic? Some of the tests I tried to do looked alright to me. Your suggestions would be valuable.

Thanks!

richardW8k: @saifsultanc `$modifiers` contains shortcode style attributes (e.g. empty="true" value="true" admin="false").  `GFCommon::get_submitted_fields` needs to be passed a comma separated string containing only the enabled modifiers (e.g. empty,value)

Idealien: > Some of the tests I tried to do looked alright to me.
I would recommend creating a form to attach to the PR with enough fields, notifications and user input steps where you can test each combination of:

- {workflow_fields:editable}
- {workflow_fields:editable,empty}
- {workflow_fields:display}
- {workflow_fields:display,empty}
- {workflow_fields:empty}
- Plus a few that also bring in the :value and :admin modifiers.

Richard has answered your other questions that probably lead to logic revisions.

saifsultanc: > @saifsultanc `$modifiers` contains shortcode style attributes (e.g. empty="true" value="true" admin="false"). `GFCommon::get_submitted_fields` needs to be passed a comma separated string containing only the enabled modifiers (e.g. empty,value)

Understood @richardW8k. Looking at the code line 73 [here](https://github.com/gravityflow/gravityflow/blob/a533d13531ea2cdfd910d2e81e4fa4edc2bf677e/includes/merge-tags/class-merge-tag-workflow-fields.php#L73), it seems we are not updating the comma-separated string (`$a`) with the values on `$modifiers`.  Only the 'default' setting of $a (lines 65-71) is used to process here (line 73)? 

In that case, we would require to update the $a based on the $modifiers (line 63) before sending it over here at code line 73? Should I go ahead and make that update as well?

richardW8k: I don't belive any changes are needed to line 73. `$this->get_attributes` handles parsing the attributes from `$modifiers` and merging them with the supplied array of defaults. That merged array is passed to `$this->get_options_string` which handles creating the comma separated string compatible with `GFCommon::get_submitted_fields`

includes/merge-tags/class-merge-tag-workflow-fields.php
```diff
@@ -109,9 +110,38 @@ public function get_options_string( $attributes ) {
 	 */
 	public function merge_tag_filter( $value, $merge_tag, $modifiers, $field ) {
 		$modifiers_array        = $field->get_modifiers();
+
+		// default case
```
stevehenty, line 114:
```suggestion
		// Default case.
```

PR #317, opened 2020-05-02
Fix Blank Form for a Partial Entry already completed

saifsultanc:
## Description
This fixes the case when a blank form appears for the partial entry step when the partial entry was already submitted.
Reference HelpScout ticket: https://secure.helpscout.net/conversation/1150735023/13425?folderId=1113492

## Testing instructions
To replicate the issue use the partial entry link to complete the entry.
Then, use the same link again and a blank form will show up.
After this update, it will show a message "This partial entry was already completed!" instead of the blank form.

N/A
 
## Screenshots
N/A

## Documentation Changes?
N/A

## Checklist:
- [x] I've tested the code.

### DIFF
--- a/includes/integrations/class-partial-entries.php
+++ b/includes/integrations/class-partial-entries.php
@@ -311,7 +311,7 @@ class Gravity_Flow_Partial_Entries {
 	/**
 	 * Returns a message to indicate that the Partial Entry was completed.
 	 *
-	 * @since 2.5
+	 * @since 2.5.10
 	 */
 	public function partial_entry_complete() {
 		return '<p>' . esc_html__( 'This entry has already been completed.', 'gravityflow' ) . '</p>';
--- a/includes/integrations/class-partial-entries.php
+++ b/includes/integrations/class-partial-entries.php
@@ -278,9 +278,7 @@ class Gravity_Flow_Partial_Entries {
 
 		// Abort if the partial entry was not found.
 		if ( empty( $entries ) ) {
-			add_filter( 'gform_form_not_found_message', function ( $message, $id ) {
-				return '<p>' . esc_html__( 'This entry has completed.', 'gravityforms' ) . '</p>';
-			}, 10, 2 );
+			add_filter( 'gform_form_not_found_message', array( $this, 'partial_entry_complete' ), 50 );
 
 			return false;
 		}
@@ -308,6 +306,15 @@ class Gravity_Flow_Partial_Entries {
 		$_POST['partial_entry_id'] = $partial_entry_id;
 
 		return $form;
+	}
+
+	/**
+	 * Returns a message to indicate that the Partial Entry was completed.
+	 *
+	 * @since 2.5
+	 */
+	public function partial_entry_complete() {
+		return '<p>' . esc_html__( 'This entry has completed.', 'gravityforms' ) . '</p>';
 	}
 
 }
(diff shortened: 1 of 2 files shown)


includes/integrations/class-partial-entries.php
```diff
@@ -278,7 +278,16 @@ public function filter_gform_pre_render( $form ) {
 
 		// Abort if the partial entry was not found.
 		if ( empty( $entries ) ) {
-			return $form;
+			$empty_form = array (
+				'title'        => $form['title'],
```
richardW8k, line 282: I suspect this will cause undefined index notices due to the id and form properties not being defined. 

Another issue to consider is that the description could be disabled on the block or shortcode used to embed the form so the message wouldn't be displayed in that scenario.

How about returning false and then using the  gform_form_not_found_message filter to return the completed message? 

The message should be located in the `esc_html__` function so it can be translated.

saifsultanc: Thanks @richardW8k. Will try to push the new update tomorrow or day after hopefully

saifsultanc: Updated @richardW8k .

PS: This was Jamie's last message to the customer: https://secure.helpscout.net/conversation/1150735023/13425?folderId=3509658#thread-3283356099

includes/integrations/class-partial-entries.php
```diff
@@ -278,7 +278,11 @@ public function filter_gform_pre_render( $form ) {
 
 		// Abort if the partial entry was not found.
 		if ( empty( $entries ) ) {
-			return $form;
+			add_filter( 'gform_form_not_found_message', function ( $message, $id ) {
```
richardW8k, line 281: I think closures require PHP 5.3 or 5.4, as we are still supporting 5.2 this would trigger an error.

Also as the message and id arguments aren't being used they could be removed.

includes/integrations/class-partial-entries.php
```diff
@@ -278,7 +278,11 @@ public function filter_gform_pre_render( $form ) {
...
+				return '<p>' . esc_html__( 'This entry has completed.', 'gravityforms' ) . '</p>';
+			}, 10, 2 );
```
richardW8k, line 283: I would increase the priority to 50 or greater to ensure our callback runs as close to last as possible.

includes/integrations/class-partial-entries.php
saifsultanc: Done the updates @richardW8k 

Do they look fine now? Any other things you'd advise?

PS: I am not sure why 'circleci' is failing for me again with the acceptance tests. It was a few weeks back as well when I realized I was missing the GF_KEY in the environment variables. When I fixed that, everything was fine. However, now this issue has recurred. I checked 'circleci' and the fails seem to be because of the GF_KEY missing. However, it is there on my env. Anything you can suggest to me to fix the issue?

includes/integrations/class-partial-entries.php
```diff
@@ -306,6 +308,15 @@ public function filter_gform_pre_render( $form ) {
 		return $form;
 	}
 
+	/**
+	 * Returns a message to indicate that the Partial Entry was completed.
+	 *
+	 * @since 2.5
+	 */
+	public function partial_entry_complete() {
+		return '<p>' . esc_html__( 'This entry has completed.', 'gravityforms' ) . '</p>';
```
richardW8k, line 317: The text domain would be gravityflow:
```suggestion
		return '<p>' . esc_html__( 'This entry has completed.', 'gravityflow' ) . '</p>';
```

PR #335, opened 2020-06-19
Inbox Paging with Show All Option

saifsultanc:
## Description
ProductBoard feature request '[Inbox Paging](https://gravityflow.productboard.com/feature-board/1199966-feature-organization/features/4382761/insights)'.
Get an option to 'Show All' entries on GravityFlow Inbox.

## Testing instructions
This update will an option of 'Show All' at the end of the Inbox table to show all of the remaining entries (after the first 150 entries).
If there are not so many entries on the Inbox, a testing possibility can be to tweak paging on the `class-api.php` temporarily. For the `get_inbox_paging` function on line 532, change the code line which has:
`'page_size' => 150,`
to a lower value (say 10):
`'page_size' => 10,`
Then, Inbox will, by default, show ten entries. On clicking 'Show All', all of the remaining entries would appear.

N/A
 
## Screenshots
N/A

## Documentation Changes?
N/A

## Checklist:
- [x] I've tested the code.


includes/pages/class-inbox.php
```diff
@@ -96,6 +104,21 @@ public static function get_defaults() {
...
 
+		$current = $_SERVER['REQUEST_URI']; 
+		$referer_parts = parse_url( $_SERVER['HTTP_REFERER'] );
+		$previous = $referer_parts['path'];
+		if ( isset( $referer_parts['query'] ) ) {
+			$previous .= '?' . $referer_parts['query'];
+		}
+
+		if ( $current != $previous ) {
+			if ( isset( $_COOKIE['more-enabled'] ) )
```
Idealien, line 115: The cookie should include some indicator it is part of Gravity Flow.
gflow-inbox-view-more perhaps?

includes/pages/class-inbox.php
```diff
@@ -96,6 +104,21 @@ public static function get_defaults() {
 			'end_date'   => '',
 		) );
 
+		$current = $_SERVER['REQUEST_URI'];
```
Idealien, line 107: One if block that has isset() checks for each of the $_SERVER params used within this block are needed.

includes/pages/class-inbox.php
```diff
@@ -104,19 +104,19 @@ public static function get_defaults() {
 			'end_date'   => '',
 		) );
 
-		$current = $_SERVER['REQUEST_URI']; 
+		$current = isset( $_SERVER['REQUEST_URI'] ) ? $_SERVER['REQUEST_URI'] : ''; 
 		$referer_parts = parse_url( $_SERVER['HTTP_REFERER'] );
 		$previous = $referer_parts['path'];
```
Idealien, line 109: There should be an else block here and/or method to appropriately mitigate should parse_url return false.
In that scenario if $previous == false skipping the rest of view more functionality.

includes/pages/class-inbox.php
```diff
@@ -96,6 +104,21 @@ public static function get_defaults() {
 			'end_date'   => '',
 		) );
 
+		$current = isset( $_SERVER['REQUEST_URI'] ) ? $_SERVER['REQUEST_URI'] : ''; 
+		$referer_parts = parse_url( $_SERVER['HTTP_REFERER'] );
```
Idealien, line 108: Accessing the admin inbox directly triggered "Notice: Undefined index: HTTP_REFERER"

includes/pages/class-inbox.php
```diff
@@ -96,6 +104,21 @@ public static function get_defaults() {
...
+		if ( isset( $referer_parts['query'] ) ) {
+			$previous .= '?' . $referer_parts['query'];
+		}
+
+		if ( $current != $previous ) {
+			if ( isset( $_COOKIE['gflow-inbox-view-more'] ) )
+				unset( $_COOKIE['gflow-inbox-view-more'] );
+		}
+		else {
+			setcookie( 'gflow-inbox-view-more', true );
```
Idealien, line 119: This may be false positive related to the other items noted above, but test attempts did encounter an error 
Warning: Cannot modify header information - headers already sent by ....wp-includes\class.wp-styles.php:287.... in 
gravityflow\includes\pages\class-inbox.php on line 119

includes/pages/class-inbox.php
```diff
@@ -81,6 +81,14 @@ public static function display( $args ) {
...
 		<?php
 		}
+
+		if ( $total_count > $page_size ) {
+			?>
+			<form method="post">
+			<input type="submit" name="view-more" value="<?php esc_html_e( 'Show All', 'gravityflow' ); ?>" />
+			</form>
+			<?php
+		}
```
Idealien, line 91: Once you select 'view all', there is currently no way to 'view less' without a page refresh.
Do you think this is something that should be possible?

includes/pages/class-inbox.php
```diff
@@ -96,6 +104,24 @@ public static function get_defaults() {
...
+				$previous .= '?' . $referer_parts['query'];
+			}
+		}
+
+		if ( $current != $previous ) {
+			if ( isset( $_COOKIE['gflow-inbox-view-more'] ) )
+				unset( $_COOKIE['gflow-inbox-view-more'] );
+		}
+		else {
+			setcookie( 'gflow-inbox-view-more', true );
```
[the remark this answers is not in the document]
saifsultanc: Set the cookie early, as you suggested. Thus, we are only working on the value updates on this block.

includes/pages/class-inbox.php
```diff
@@ -81,6 +81,14 @@ public static function display( $args ) {
 			</div>
 		<?php
 		}
+
+		if ( $total_count > $page_size ) {
```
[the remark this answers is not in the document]
saifsultanc: Initialized the variable early to prevent an undefined error.

includes/pages/class-inbox.php
saifsultanc: Maybe, the View Less can be worked on as an additional feature. Once everything is sorted on this PR? For now, if we move away from the inbox page and come back to it later, the display is reset to the default paging size.

richardW8k:
@saifsultanc can you merge master in here and fix the undefined index view_more error which is causing the tests to fail: https://travis-ci.com/github/gravityflow/gravityflow/jobs/454915287

Idealien:
@saifsultanc - Is this ready for re-review now?

speerface:
I was thinking about this a bit over the last week, and I'm curious if we considered using `user_meta` instead of a cookie to store this preference? I feel like that might gel a bit more with how a lot of other Dashboard UI stuff works in WordPress (for instance, the order of metaboxes, or the `screen` options). Functionally it'd wind up being close to the same, but it'd be user-based instead of browser-based. Any thoughts?

Idealien:
> I'm curious if we considered using user_meta instead of a cookie to store this preference?
Not to my knowledge, but that would preclude email token/field based assignees from accessing the feature. However, the email assignee use case generally involves less total entries where view more/less is as valuable.

stevehenty:
Closing in favour of https://github.com/gravityflow/backlog/issues/78

PR #336, opened 2020-06-19
Date fields conditions on User Input Step dependent on other Date fields

saifsultanc:
## Description
HelpScout# [13891](https://secure.helpscout.net/conversation/1194802467/13891?folderId=1113492#thread-3412151553)

Selecting a Date Field on User Input dependent on a non-editable Date Field with conditions from GP-Limit-Dates did not work.

## Testing instructions
Testing form from the ticket, and other possible scenarios where a Date field on User Input step is dependent on a non-editable Date field.

N/A
 
## Screenshots
N/A

## Documentation Changes?
N/A

## Checklist:
- [x] I've tested the code.

### DIFF
--- a/includes/pages/class-entry-editor.php
+++ b/includes/pages/class-entry-editor.php
@@ -107,6 +107,15 @@ class Gravity_Flow_Entry_Editor {
 	private $_requires_pricing_inputs = false;
 
 	/**
+	 * Indicates if there is an editable date field which requires the presence of other date fields to function.
+	 *
+	 * @since 2.5.12
+	 *
+	 * @var bool
+	 */
+	private $_requires_date_inputs = false;	
+
+	/**
 	 * Gravity_Flow_Entry_Editor constructor.
 	 *
 	 * @param array             $form                 The current form.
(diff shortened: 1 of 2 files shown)


includes/pages/class-entry-editor.php
```diff
@@ -368,6 +382,44 @@ public function is_pricing_field_required( $field ) {
...
+			if ( $field['gpLimitDatesminDateMod'] != '' ) {
+				return true;
+			}
+		}
+
+		if ( isset( $field['gpLimitDatesmaxDateMod'] ) ) {
+			if ( $field['gpLimitDatesmaxDateMod'] != '' ) {
+				return true;
+			}
+		}
```
richardW8k, line 405: These should use object notation. While array access is currently supported it's not recommended as it will most likely be removed in a future GF release.

From what I understand about GP Limit Dates `gpLimitDatesminDate` and `gpLimitDatesmaxDate` are probably the best properties to check as they will contain the IDs of the fields the limit is based on. They can also contain other values such as `_custom_` so an `is_numeric()` check might be needed to determine if the value is a field ID e.g. 

`! empty( $field->gpLimitDatesminDate ) && is_numeric( $field->gpLimitDatesminDate )`

I wonder if it should take a similar approach to the calculation field dependencies? Store the field IDs from the `gpLimitDatesminDate` and `gpLimitDatesmaxDate` properties in an array and then check that when `can_remove_field()` is used to see if the field can be removed.

PR #353, opened 2020-09-04
Fix Fatal Error with Assignees on User Input Step

saifsultanc:
## Description
Re: [Issue#41](https://github.com/gravityflow/backlog/issues/41)

## Testing instructions
1. Create a form with a User Input step and select Step Assignees, and try to save step settings.
(Alternatively can use the form export attached on the Issue page).
2. Saving the step settings will result in a fatal error.
3. Switch to this branch, and then repeat step 1.
4. Step settings success.
5. Test the workflow to confirm everything works as expected.

N/A
 
## Screenshots
N/A

## Documentation Changes?
N/A

## Checklist:
- [x] I've tested the code.

### DIFF
--- a/change_log.txt
+++ b/change_log.txt
@@ -1,0 +1,1 @@
+- Fix an issue with Step Assignees on User Input Step causing a Fatal Error.
--- a/includes/steps/class-step.php
+++ b/includes/steps/class-step.php
@@ -1616,13 +1616,15 @@ abstract class Gravity_Flow_Step extends stdClass {
 		$has_editable_fields = ! empty( $this->editable_fields );
 
 		foreach ( $this->assignees as $assignee_key ) {
-			$args = $this->get_assignee_args( $assignee_key );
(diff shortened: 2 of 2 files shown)


richardW8k (changes requested):
I wasn't able to replicate the fatal error with master and Gravity Forms 2.4.20 using my test forms, a new form, or the form from the issue zip.

Inspecting the json from the issue zip I found the reason for the issue, there is an extra item in the array with an empty key and empty value: https://d.pr/i/bzdgUy. 

When I tested that empty item is removed when the step settings are rendered so it is not present when the step is saved.

I'm not sure how that ended up in the array but the way I would prevent it causing an issue would be to make a change to the foreach in `Gravity_Flow_Step::maybe_add_select_assignees()` so it skips processing the `$assignee_key` if it is empty.

richardW8k (approved):

PR #354, opened 2020-09-10
Added filter to remove inbox count

saifsultanc:
## Description
Re: [HS#14662](https://secure.helpscout.net/conversation/1274562929/14662/) to remove inbox count red bubble.

## Testing instructions
- After switching to this branch, add the filter call on functions.php:
`add_filter( 'gravityflow_inbox_count_display', '__return_false', 1 );`
- Refresh Wordpress Dashboard and confirm the filter count is no longer.
- Remove the filter call from functions.php, and check to see the inbox count value is back.

N/A
 
## Screenshots
N/A

## Documentation Changes?
Add documentation for the filter `gravityflow_inbox_count_display`

`add_filter( 'gravityflow_inbox_count_display', '__return_false', 1 );`

Alternativey,
```
add_filter( 'gravityflow_inbox_count_display', 'sh_gravityflow_inbox_count_display', 10, 1 );
function sh_gravityflow_inbox_count_display( $show ) {
	return false;
}
```

## Checklist:
- [x] I've tested the code.

### DIFF
--- a/class-gravity-flow.php
+++ b/class-gravity-flow.php
@@ -5989,6 +5989,12 @@ jQuery('#setting-entry-filter-{$name}').gfFilterUI({$filter_settings_json}, {$va
 		 * @param array $menu The current WP Dashboard Menu.
 		 */		
 		public function show_inbox_count( $menu ) {
+
+			$show = apply_filters( 'gravityflow_show_inbox_count', $show );
+			if ( isset( $show ) && ! $show ) {
+				return $menu;
+			}
+
 			$custom_labels = get_option( 'gravityflow_app_settings_labels', array() );
 			$custom_navigation_labels = rgar( $custom_labels, 'navigation' );
 			$custom_workflow_label = rgar( $custom_navigation_labels, 'workflow' );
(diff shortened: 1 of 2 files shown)


Idealien (changes requested):
Functionality is sound.
I'd recommend the filter be renamed to gravityflow_inbox_count_display to be consistent with the [naming conventions](https://github.com/gravityflow/gravityflow/wiki/Naming-conventions-for-hooks) to give an emphasis to the object being filtered, followed by when it is going to fire.

class-gravity-flow.php
```diff
@@ -5989,6 +5989,12 @@ public function filter_after_update_entry( $form, $entry_id ) {
 		 * @param array $menu The current WP Dashboard Menu.
 		 */		
 		public function show_inbox_count( $menu ) {
+
+			$show = apply_filters( 'gravityflow_inbox_count_display', $show );
+			if ( isset( $show ) && ! $show ) {
```
richardW8k, line 5994: We don't usually use an approach like this when adding filters. Normally the argument would be defaulted to `true` or `false` e.g.

```suggestion
			$show = apply_filters( 'gravityflow_inbox_count_display', true );
			if ( ! $show ) {
```

There should also be a docblock with since and param tags.

richardW8k (changes requested):

richardW8k (approved):

PR #356, opened 2020-10-01
Add CSS ready class to entry detail page

saifsultanc:
## Description
Re: [HS#14499](https://secure.helpscout.net/conversation/1260621046/14499?folderId=3509661#thread-3635195004)
The issue about not being able to style entry detail view. Adding CSS ready classes for the user customization.

## Testing instructions
Test with the entry detail view to verify the table rows have class `entry-view-section-break`.

N/A
 
## Screenshots
N/A

## Documentation Changes?
N/A

## Checklist:
- [x] I've tested the code.

### DIFF
--- a/change_log.txt
+++ b/change_log.txt
@@ -1,2 +1,3 @@
 - Fix an issue with Step Assignees on User Input Step causing a Fatal Error.
 - Added filter 'gravityflow_inbox_count_display' for customization of the Gravity Flow Inbox. The Inbox count value can be disabled using it.
+- Added CSS ready classes to entry detail page.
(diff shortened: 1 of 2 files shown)


Idealien (changes requested):
The class entry-view-section-break has specific association in existing code to the section field type.
Please modify so that the row class being added is entry-view-row.

How does this provide any more specificity for CSS purposes given every row has the same class?  With browsers still not supporting the [:has selector](https://caniuse.com/css-has) a jQuery selector relating to parent rows of
$("td.entry-view-field-value").parents("tr")...
is negligibly different than
$("td.entry-view-field-value").parents("tr.entry-view-row")

Should this be looking at mapping some/all of a fields' custom CSS classes (where actual CSS ready classes defined) onto the tr/td to more generally support the [CSSReady classes](https://docs.gravityforms.com/css-ready-classes/) on entry detail as the user input does today?

Idealien (changes requested):
Please see/respond to the questions from https://github.com/gravityflow/gravityflow/pull/356#pullrequestreview-501634583 regarding the larger context/approach the PR.

PR #359, opened 2020-10-08
Fix workflow showing same step on submission for email based assignee

saifsultanc:
## Description
Re: [HS# 14496](https://secure.helpscout.net/conversation/1277212685/14687?folderId=1113497#thread-3671826413)
An issue with the email assignee not reflecting approval (or move to the next step) on the Workflow Inbox front-end display.

## Testing instructions
1. Create a form with an email field, and any other field (say discussion)
2. Create a user input step with the email field as assignee, passing the workflow entry link merge tag via email.
3. Check the issue is resolved when working through the user input step.

N/A
 
## Screenshots
N/A

## Documentation Changes?
N/A

## Checklist:
- [x] I've tested the code.

### DIFF
--- a/change_log.txt
+++ b/change_log.txt
@@ -1,1 +1,2 @@
 - Updated Inbox count to be disabled by default, can be enabled with the filter 'gravityflow_inbox_count_display'.
+- Fixed an issue where the front-end workflow was showing stuck on same step for email based assignee.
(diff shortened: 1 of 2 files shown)


Idealien (changes requested):
This has unintended consequences that solution should be more granular?

**Additional Test Case:**
- Create form with input fields
- Add 2 approval steps with different assignees (user 1, user 2)
- On master branch - when step 1 completes user 1 can see that next step is assigned to User 2.
- On PR branch, user 1 just sees that the workflow is 'pending'

Idealien:
- Without this code change as is - email field based assignees see a 2nd entry screen after submit but cannot process/action.
- Issue with this code change as is - non-email field based user no longer see the next step assignees (when they are not next step assignee)

The issue is secondary in comparison to what it resolves, and also some discussion to be had whether the remaining issue is actually how the assignee display should be handled. Approving the PR as is and we can address via additional fix as required.

Idealien (approved):
Please follow-up the discussion regarding next step assignee detail display in our dev channel to confirm if a subsequent PR is required what solution form may take.

PR #363, opened 2020-10-23
Fix gravityview entry revision user input

saifsultanc:
## Description
Re: [HS#14607](https://secure.helpscout.net/conversation/1270186867/14607?folderId=1113492#thread-3788450740)
With Gravity View Entry Revisions, a successfully completed User Input Step on the workflow processes it's expiration step (when it should not).

## Testing instructions
1. Replicate the form and workflow settings as on the sandbox provided. Ensure Gravity View and Entry Revisions are enabled.
2. After the first step "Need Admin Response", the expiration step email is erroneously triggered.
3. Switch to this branch and repeat 2. This time, the erroneous email trigger doesn't happen,

With this update, entry revisions for User Input step will not be stored. Maybe this limitation or exception with Gravity View Entry Revisions can be updated [here](https://docs.gravityflow.io/article/30-integration-with-gravity-forms-add-ons) for the time being.

N/A
 
## Screenshots
N/A

## Documentation Changes?
N/A

## Checklist:
- [x] I've tested the code.

### DIFF
--- a/change_log.txt
+++ b/change_log.txt
@@ -5,3 +5,4 @@
 - Fixed an issue with the print button on workflow details page.
 - Fixed an issue with the print button on front-end pages.
 - Fix JS errors with multiple status shortcodes on the same page.
+- Fix an issue with Gravity View Entry Revisions causing Gravity Flow to run steps out of sync.
--- a/change_log.txt
+++ b/change_log.txt
@@ -5,4 +5,4 @@
 - Fixed an issue with the print button on workflow details page.
 - Fixed an issue with the print button on front-end pages.
 - Fix JS errors with multiple status shortcodes on the same page.
-- Fix an issue with Gravity View Entry Revisions causing Gravity Flow to run steps out of sync.
+- Fixed an issue where workflow steps are processed for entry revisions created by GravityView Entry Revisions.
--- a/includes/steps/class-step-user-input.php
+++ b/includes/steps/class-step-user-input.php
@@ -1131,6 +1131,8 @@ class Gravity_Flow_Step_User_Input extends Gravity_Flow_Step {
 	public function save_entry( $form, &$entry, $editable_fields ) {
 	    global $wpdb;
 
+		remove_action( 'gform_after_update_entry', array( GV_Entry_Revisions::get_instance(), 'gform_after_update_entry' ), -100 );
+		
 		$this->log_debug( __METHOD__ . '(): Saving entry.' );
 
 		$is_new_lead = $entry == null;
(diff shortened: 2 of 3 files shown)


includes/steps/class-step-user-input.php
```diff
@@ -1131,6 +1131,8 @@ public function entry_detail_status_box( $form ) {
 	public function save_entry( $form, &$entry, $editable_fields ) {
 	    global $wpdb;
 
+		remove_action( 'gform_after_update_entry', array( GV_Entry_Revisions::get_instance(), 'gform_after_update_entry' ), -100 );
```
richardW8k, line 1134: This will trigger a fatal error if GV Entry Revisions is not active.

richardW8k (changes requested):
When I tested this I found that the email is not being sent for the current entry, it is actually being sent for the revision which GV Entry Revisions is creating using `GFAPI::add_entry()`.

That method triggers the `gform_post_add_entry` hook which we use to start a workflow for the new entry. 

It looks like GV sets the `status` of the entry being created to `gv-revision` so what we could do is update the first if statement in `Gravity_Flow::action_gform_post_add_entry()` so it also returns early if `rgar( $entry, 'status' ) !== 'active'`. That will prevent the workflow running for the revisions.

change_log.txt
```diff
@@ -4,4 +4,5 @@
 - Updated Inbox count to be disabled by default, can be enabled with the filter 'gravityflow_inbox_count_display'.
 - Fixed an issue with the print button on workflow details page.
 - Fixed an issue with the print button on front-end pages.
-- Fix JS errors with multiple status shortcodes on the same page.
\ No newline at end of file
+- Fix JS errors with multiple status shortcodes on the same page.
+- Fix an issue with Gravity View Entry Revisions causing Gravity Flow to run steps out of sync.
```
richardW8k, line 8:
```suggestion
- Fixed an issue where workflow steps are processed for entry revisions created by GravityView Entry Revisions.
```

includes/steps/class-step-user-input.php
```diff
@@ -1130,7 +1130,7 @@ public function entry_detail_status_box( $form ) {
 	 */
 	public function save_entry( $form, &$entry, $editable_fields ) {
 	    global $wpdb;
-
+
```
richardW8k, line 1133: This indent should be removed

richardW8k (approved):
Good to merge once the change log is updated and the indent is removed from the changed line in the user input file.

PR #366, opened 2020-11-06
Fix Step Condition having Form Submission Step fetching wrong entry meta

saifsultanc:
## Description
Re: [HS# 15322](https://secure.helpscout.net/conversation/1325992575/15322?folderId=1113492#thread-3825730334)
When processing the step condition having a Form Submission step, the conditional logic wasn't evaluating properly, thus failing the entire workflow. The issue was the `$form_id` on `get_feed_condition_entry_meta()` was taken as the child `$form_id`. To fix this, we check if it's a `$parent_entry` and get the parent `$form_id`.

## Testing instructions
1. Use the testing form provided on the ticket.
2. Replicate the issue with `master`.
3. Switch to this branch and see the issue is resolved.
4. Test for other parent-child scenarios with the Form Connector plugin, like New Entry, etc.

N/A
 
## Screenshots
N/A

## Documentation Changes?
N/A

## Checklist:
- [x] I've tested the code.

### DIFF
--- a/class-gravity-flow.php
+++ b/class-gravity-flow.php
@@ -8710,9 +8710,8 @@ AND m.meta_value='queued'";
 				return true;
 			}
 
-			$form_id         = $form['id'];
-			$step_id         = $this->get_current_feed_id();			
-			$entry_meta      = array_merge( $this->get_feed_condition_entry_meta( $form_id, $step_id ), $this->get_feed_condition_entry_properties() );
+			$form_id         = $form['id'];			
+			$entry_meta      = array_merge( $this->get_feed_condition_entry_meta( $form_id ), $this->get_feed_condition_entry_properties() );
 			$entry_meta_keys = array_keys( $entry_meta );
 			$match_count     = 0;
 
--- a/class-gravity-flow.php
+++ b/class-gravity-flow.php
@@ -8552,6 +8552,13 @@ AND m.meta_value='queued'";
 		public function get_feed_condition_entry_meta() {
 			$step_id    = absint( rgget( 'fid' ) );
 			$form_id    = absint( rgget( 'id' ) );
+			
+			if ( isset( $_POST['workflow_parent_entry_id'] ) ) {
+				$parent_entry_id = absint( rgpost( 'workflow_parent_entry_id' ) );
+				$parent_entry    = GFAPI::get_entry( $parent_entry_id );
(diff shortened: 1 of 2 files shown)


richardW8k (changes requested):
I'd rather not add Form Connector specific code here. How about updating `get_feed_condition_entry_meta()` to include `$form_id` and `$step_id` as arguments and then update `settings_feed_condition()` and `evaluate_conditional_logic()` to pass them?

class-gravity-flow.php
```diff
@@ -8545,13 +8547,14 @@ public function settings_feed_condition( $field, $echo = true ) {
 		/**
 		 * Get the entry meta for use with the feed_condition setting.
 		 *
-		 * @since 1.7.1-dev
+		 * @since 2.6
```
richardW8k, line 8550: The 1.7.1 since tag should remain.
The 2.6 since tag should have a description indicating the two new args were added.

class-gravity-flow.php
```diff
@@ -8706,7 +8709,9 @@ public function evaluate_conditional_logic( $logic, $form, $entry ) {
 				return true;
 			}
 
-			$entry_meta      = array_merge( $this->get_feed_condition_entry_meta(), $this->get_feed_condition_entry_properties() );
+			$form_id         = $form['id'];
+			$step_id         = $this->get_current_feed_id();
```
richardW8k, line 8713: You can actually remove this line as the step id is only needed on the step configuration page to prevent them configuring a rule based on the current step.

class-gravity-flow.php
```diff
@@ -8545,13 +8547,14 @@ public function settings_feed_condition( $field, $echo = true ) {
...
+		 *
+		 * @param int $form_id The form ID.
+		 * @param int $step_id The step ID.		 
 		 *
 		 * @return array
 		 */
-		public function get_feed_condition_entry_meta() {
-			$step_id    = absint( rgget( 'fid' ) );
-			$form_id    = absint( rgget( 'id' ) );
+		public function get_feed_condition_entry_meta( $form_id, $step_id ) {
```
richardW8k, line 8557: Since these are new args we should probably set defaults (`$form_id = 0, $step_id = 0`) in case anyone is using this function in their own code.

richardW8k (changes requested):

richardW8k (approved):

PR #367, opened 2020-11-11
Feature/status entry url filter

Idealien:
[#HS 14297](https://secure.helpscout.net/conversation/1333659385/15380/#)

## Description
The status view lacked a filter to customize what URL the user would be sent to when displayed via the shortcode. Similar to gravityflow_entry_url_inbox_table for the inbox view.

## Testing instructions
- Create a page with a [gravityflow page="status"] shortcode on it.
- Add the following snippet to your functions.php file and adjust the page / get_the_ID() control for your environment.
```
add_action( 'gravityflow_entry_url_status_table', 'jo_status_entry_url_blank', 10, 4 );
function jo_status_entry_url_blank( $entry_url, $form_id, $entry_id, $entry ) {
	if ( ! get_the_ID() == '285' ) {
		return $entry_url;
	} else {
		return '#';
	}  
}
```
- Verify that the link can be clicked but user does not navigate away from the status list view.
 
## Documentation Changes?
Yes - Draft entry for gravityflow_entry_url_status_table has been prepared in Developers / Actions & Filters category.

## Checklist:
- [X] I've tested the code.
- [X] My code follows the WordPress code style. 
- [X] My code follows the accessibility standards. 
- [X] My code follows the inline documentation standards.


includes/pages/class-status.php
```diff
@@ -1127,7 +1127,18 @@ public function column_payment_status( $item ) {
...
+		$entry_url = sprintf( '%s&id=%d&lid=%d', $this->detail_base_url, $entry['form_id'], $entry['id'] );
+		/**
+		 * Allows the entry url to be filtered in the status table.
+		 *
+		 * @since 2.6.1
+		 *
+		 * @param string $entry_url   The url to be linked to.
+		 * @param int    $form_id     The Form ID.
+		 * @param int    $entry_id    The Entry ID.
+		 * @param array  $item        The entry array.
```
richardW8k, line 1139: I'd go with `$entry` instead of `$item` here. Other than that it's good to merge.

richardW8k (approved):

includes/pages/class-status.php
Idealien: Updated param.

PR #368, opened 2020-11-14
Added "gravityflow_routing_values_markup" filter to allow filtering t…

spivurno:
…he markup rendered for the Conditional Routing's value select.

## Testing instructions

Here's how we'll be using this in [Populate Anything](https://gravitywiz.com/documentation/gravity-forms-populate-anything) to allow users to specify custom values for dynamically populated choice-fields.

```js
gform.addFilter( 'gravityflow_routing_values_markup', function( markup, filter, selectedOperator, index ) {

	if ( ! filter || ! jQuery.isNumeric( filter.key ) ) {
		return markup;
	}

	var field = GetFieldById( filter.key );
	if ( ! field ) {
		return markup;
	}

	if ( field['gppa-choices-enabled'] ) {
		markup = '<input type="text" value="" class="gform-routing-value value_{0}" />'.format( index );
	}

	return markup;
} );
```

## Checklist:
- [x] I've tested the code.

### DIFF
--- a/js/routing-setting.js
+++ b/js/routing-setting.js
@@ -242,7 +242,17 @@
                 str = '<input type="text" value="" class="gform-routing-value value_{0}" />'.format(index);
             }
 
-            return str;
+            /**
+             * Filter the markup rendered for the Conditional Routing's value select.
+             *
+             * @param str    str              The markup to be rendered.
+             * @param object filter           An object containing the properties of the current condition.
+             * @param str    selectedOperator The operator that is selected for the current condition.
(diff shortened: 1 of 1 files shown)


davidzuluaga5:
Many thanks! I really need this for real use case with clients.

richardW8k (approved):

stevehenty:
I've [reverted](https://github.com/gravityflow/gravityflow/pull/372) this for the moment until we can come up with a better way to handle it.

PR #382, opened 2021-01-31
Update entry filter setting to support the filters being passed in the field config

richardW8k:
## Description
The way the entry filter setting currently works is that it uses the form id which is passed in the field config to call `GFCommon::get_field_filter_settings()`. This is fine for local forms but when the form is located on another site this results in the setting being populated with local data which in most cases won't match the selected form. 

## Testing instructions
- See https://github.com/gravityflow/gravityflowformconnector/pull/43
- On the Update Fields step confirm that the setting displays the fields, entry meta, and users from the remote site and not a local form.

N/A

## Documentation Changes
N/A

## Checklist:
- [x] I've tested the code.


saifsultanc (approved):

Idealien (approved):
Added change log entry based on PR notes.

stevehenty:
@richardW8k Is this ready to merge?

PR #385, opened 2021-02-14
Master fix 67 update assignee email message to use 2.5 editor

speerface:
## Description
The Message textarea for the Assignee Email in a Step was using the soon-to-be-deprecated 2.4 version of the editor. In Gravity Forms 2.5, the process for defining Settings Fields updated to move away from `visual_editor` and to a standardized `textarea` type. 

This fix simply updates the properties we send to the Settings API to use the `textarea` type along with `use_editor` to ensure it remains a WYSIWYG field. 

Additionally, Gravity Forms is expecting each field to be wrapped with a `div` with a combination of classes (`gform-settings-field gform-settings-field__{input type}` in order to inherit the correct styles. In this case, since this was a "tabbed" field in our UI, we were only setting the `gravityflow-tab-field` class on the field wrapper in our custom logic. This fix updates things to also add the additional `gform-settings-field` classes to our markup so that the correct style inheritance applies. 

Resolves issue 67: https://github.com/gravityflow/backlog/issues/67

## Testing instructions
1. Set up a form and add some Workflow Steps
2. Edit a Step, and scroll down to the Message field for the Assignee
3. Ensure that in Gravity Forms 2.5, the display matches the screenshot below. 
4. Switch to Gravity Forms 2.4, and ensure that the display remains as it was before this change (these changes should only fundamentally impact 2.5+). 

 
## Screenshots

**The correct display in GF 2.5**
correct-layout

## Checklist:
- [x] I've tested the code.

### DIFF
--- a/change_log.txt
+++ b/change_log.txt
@@ -1,4 +1,4 @@
 - Added action gravityflow_step_start for custom logic to perform when any step starts.
 - Fixed an issue that prevented use of gravityflow_status_filter with an 'any' mode selection from being applied correctly.
 - Updated the Status Page to use a CSS Loading Spinner to replace the deprecated Spinner Gif from Gravity Forms Core.
-- Updated the Assignee Email Message Textarea to use an updated version of the TinyMCE Editor from Gravity Forms 2.5
+- Updated the Assignee Email Message Textarea to improve support for the TinyMCE Editor from Gravity Forms 2.5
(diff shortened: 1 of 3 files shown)


Idealien (approved):
Functionality works as expected with Gravity Forms 2.4 and 2.5. Review approved - please revise the change log message slightly to indicate improved support for 2.5. As written, it could be interpreted that it has a dependency on 2.5 for the text area to function.

richardW8k (approved):

PR #390, opened 2021-03-19
Fix conflict with Feed Add-On Framework feed ordering functionality

travislopes:
## Description
This PR fixes a conflict that occurs due to how Gravity Flow forks the feed ordering method implemented in the Feed Add-On Framework.

Since Gravity Flow uses a custom action name but the same Feed Add-On Framework method name, I've updated the method to confirm the action is for Gravity Flow feed ordering before continuing processing.

I also added a nonce check for added security.

## Testing instructions
1. Create a form with at least two workflow steps.
2. On the Workflow Steps page, re-order the feeds.
   - The AJAX response should be `[["ok"],200]`.
   - When you reload the page, the order should match the new feed order.

## Checklist:
- [x] I've tested the code.

### DIFF
--- a/class-gravity-flow.php
+++ b/class-gravity-flow.php
@@ -2077,6 +2077,10 @@ PRIMARY KEY  (id)
 		 * Ajax handler for the request to save the custom feed order.
 		 */
 		public function ajax_save_feed_order() {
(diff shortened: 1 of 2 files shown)


stevehenty:
Thanks @travislopes !

richardW8k (approved):

PR #391, opened 2021-03-23
Fix Order Summary always displaying with Complete Step Settings

saifsultanc:
## Description
Addresses [Issue #86](https://github.com/gravityflow/backlog/issues/86)
Order Summary always displays at end of Workflow, regardless of whether it is enabled or disabled on the Complete Step Settings.

## Testing instructions
1. Create a form with Pricing Fields.
2. Add a few workflow steps (like Approval, User Input, etc).
3. Edit Complete Step settings to disable Order Summary.
4. Process the form and workflow, at the end of the workflow the Order Summary would still be displayed.
5. Now switch to this branch, repeat step 4 and now Order Summary will not be displayed.
6. Update workflow to enable Order Summary on Complete Step and repeat step 4 again. Now Order Summary will display, as expected.

Added test `0052SummaryCompleteStepCept.php` to check for Order Summary Settings on Complete Step
 
## Screenshots
N/A

## Documentation Changes
N/A

## Checklist:
- [x] I've tested the code.


includes/pages/class-entry-detail.php
```diff
@@ -773,6 +773,11 @@ public static function entry_editor( $form, $entry, $current_step, $display_empt
 	 */
 	public static function maybe_show_products_summary( $form, $entry, $current_step ) {
 		$summary_enabled = true;
+		$complete_step = gravity_flow()->get_workflow_complete_step( $form['id'] );
+		if ( ! $current_step && $complete_step ) {
+			$current_step = $complete_step;
```
Idealien, line 778: While this approach is functional, it modifies $current_step in addition to determining the value of $summary_enabled. Please put the additional check of feed/setting inside an else/if block so that current / complete are evaluated separately. It reduces risk that future enhancements evaluating step do not get the expected value.

Idealien (changes requested):

includes/pages/class-entry-detail.php
Idealien: We should also check whether existing acceptance test(s) validate this setting or could be enhanced to as a part of this PR.

saifsultanc: @Idealien : I had initially done it that way, but to avoid code duplication/repeated logic went for a workaround like this. Would it make sense to use a different variable like `$processing_step`, and then use it as :

$complete_step = gravity_flow()->get_workflow_complete_step( $form['id'] );
$processing_step = ( ! $current_step && $complete_step ) ? $complete_step : $current_step;

and now proceed with the rest of the logic on `$processing_step`?

Idealien: Good idea, yes.

Idealien (approved):

PR #393, opened 2021-04-02
Fix HTML Fields on Workflow Detail Page

saifsultanc:
## Description
Re: [Issue#85](https://github.com/gravityflow/backlog/issues/85)
HTML fields with conditional logic display regardless of whether the condition is true or false on Workflow detail.

## Testing instructions
1. Create a form with a text field and say two HTML fields A and B.
2. Set field A to be displayed only if the text field is empty.
3. Set field B to be displayed only if the text field is not empty.
4. Add an Approval Workflow Step.
5. Process the form and the workflow step.
6. At the end of the workflow, we will be seeing both the HTML fields (instead of only one).
7. Update to this branch, and now we will only see HTML field A when the text field is empty. When the text field is not empty, we will only see HTML field B.

Added test `0053ConditionalHTMLWorkflowDetailCept.php` to check for conditional HTML display on Workflow Details page.
 
## Screenshots
N/A

## Documentation Changes
N/A

## Checklist:
- [x] I've tested the code.


Idealien (changes requested):
The code change does affect the change required. However, I wonder if the approach taken is either introducing unintended impacts or needs additional work? 

As an example - based on the test scenario of paragraph field + 2 html fields. An approval step with 'display all fields' will display 1 html field whereas a step with 'display all fields except' that excludes the paragraph field will still show both html fields.

saifsultanc:
> The code change does affect the change required. However, I wonder if the approach taken is either introducing unintended impacts or needs additional work?
> 
> As an example - based on the test scenario of paragraph field + 2 html fields. An approval step with 'display all fields' will display 1 html field whereas a step with 'display all fields except' that excludes the paragraph field will still show both html fields.

While I did ponder on an alternate approach like some specific logic to cater only to HTML fields, but then an assessment of the if-conditionals made me visualize that we had a more general if- conditional preceding the more specific elseif- one. It made sense to have the specific one first.

The case you mentioned for testing doesn't produce that behavior. Even with the 'display all fields except' I see only the HTML field that is expected to appear and not the other. Please let me know if there is any other setting for this that made you note this issue. Thanks!

Idealien (approved):
The HTML fields are following consistent pattern for how the rest of field types respect conditional logic as precedence over display field settings. If users need to override they have https://docs.gravityflow.io/article/44-gravityflowworkflowdetaildisplayfield.

Approved. Thx Saif.

PR #398, opened 2021-04-12
Patch issue with inbox chunk loading on all screens

faction23:
## Description
The inbox chunk I have started to work on that is injected by webpack was loading on all screens when it shouldn't have due to an incorrect exists check. This was noticed as well in console as an error with minified files since the chunk path in the grunt build was not set. We will resolve the latter in subsequent work, the former will cause it to not occur for now as the exists check will never trigger the chunk load.

## Testing instructions

* Load up the admin and any Gravity Flow screen.
* Confirm the message "'Gravity Flow Common: Initialized inbox component.'" is not present in the console.

## Checklist:
- [x] I've tested the code.


saifsultanc (approved):
That looks good Sam. Does the job for now.

PR #416, opened 2021-07-15
Fix missing extension icons under Forms Toolbar Settings

saifsultanc:
## Description
This fixes the missing extension icons which are displayed under the Forms Toolbar Settings.

This resolves [gravityflow/backlog#126](https://github.com/gravityflow/backlog/issues/126).

## Testing instructions
1. Create a simple form.
2. Hover on the Toolbar Settings in the form
3. Note many Gravity Flow Extension icons are missing (blank).
4. Switch to this branch.
5. Repeat Step 2 and confirm the icons now display fine.

N/A
 
## Screenshots
N/A

## Documentation Changes
N/A

## Checklist:
- [x] I've tested the code.


Idealien (changes requested):
This does fix majority of extensions. The function should be hooked later, perhaps at 20, to ensure it applies for all our extensions. PDF was missed based on my testing. Also requires change log note.

richardW8k (commented):
Have you tried overriding `get_menu_icon()` instead of using the `gform_form_settings_menu` filter?

```
	public function get_menu_icon() {
		return 'dashicons-gravityflow-icon';
	}
```

saifsultanc:
> Have you tried overriding `get_menu_icon()` instead of using the `gform_form_settings_menu` filter?
> 
> ```
> 	public function get_menu_icon() {
> 		return 'dashicons-gravityflow-icon';
> 	}
> ```

@richardW8k It doesn't load the icon for all the extensions. On my testing (after overriding), it only worked with Incoming Webhook.

I took @Idealien's advice on the priority, and that works just fine.

richardW8k:
@saifsultanc did you also add `get_menu_icon()` to `Gravity_Flow_Feed_Extension`?

saifsultanc:
> @saifsultanc did you also add `get_menu_icon()` to `Gravity_Flow_Feed_Extension`?

Yes @richardW8k did that to both (extension and feed extension).

richardW8k:
@saifsultanc i figured out why, those add-ons are overriding `add_form_settings_menu()` so  they don't call `get_menu_icon()` which is new in 2.5.  We should still add `get_menu_icon()` as it resolves the issue where the icons are missisng from the Forms > Settings > Uninstall page.

saifsultanc:
> @saifsultanc i figured out why, those add-ons are overriding `add_form_settings_menu()` so they don't call `get_menu_icon()` which is new in 2.5. We should still add `get_menu_icon()` as it resolves the issue where the icons are missisng from the Forms > Settings > Uninstall page.

Done that @richardW8k. Also, added `get_menu_icon()` for the `Gravity_Flow` class.

For the Workflow > Settings page added the update [here](https://github.com/gravityflow/gravityflow/blob/master-fix-126-extension-icons-on-settings/class-gravity-flow.php#L9363). We should also update the icons here, or just leave them to the generic?

richardW8k (changes requested):
This works but i've noticed the filter is running for every active extension even though the first extension to add the filter resolves the issue for them all. 

How about moving the filter to `Gravity_Flow` instead so it only runs once and adjusting the if statement to the following so it won't apply to any extensions we decide to add custom icons to in future releases?

```
if ( empty( $menu_item['icon'] ) && strpos( $menu_item['name'], 'gravityflow' ) === 0 ) {
```

richardW8k (approved):

PR #419, opened 2021-07-30
Add Gravity Forms Uninstall Page Look to Gravity Flow Uninstall Page

saifsultanc:
## Description
The Gravity Forms uninstall page displays uninstall options for Gravity Forms and all of its add-ons in a clean tabular format. On the other hand, the Gravity Flow uninstall page only has the core Gravity Flow uninstall option with no information about its extensions.

This PR adds the Gravity Forms uninstall page look to the Gravity Flow uninstall page. All currently active Gravity Flow extensions are displayed on the uninstall page, similar to how Gravity Forms displays its addons on the uninstall page.

This resolves [gravityflow/backlog#111](https://github.com/gravityflow/backlog/issues/111).

## Testing instructions
Compare the Gravity Flow uninstall page after updating to this branch with the current Gravity Forms uninstall page.

N/A
 
## Screenshots
N/A

## Documentation Changes
N/A

## Checklist:
- [x] I've tested the code.


class-gravity-flow.php
```diff
@@ -9590,5 +9590,82 @@ public function action_update_site_option_auto_update_plugins( $option, $value,
...
+		 * @since  2.5
+		 * @return array
+		 *
+		 */
+		private static function uninstall_extensions() {
+
+			$installed_extensions = GFAddOn::get_registered_addons( true );
+			$installed_extensions = array_filter( $installed_extensions, function( $extension ) {
+				return substr( $extension->get_slug(), 0, 11 ) === 'gravityflow';
+			});
```
richardW8k, line 9645: Support for the true arg was only added in 2.5.6, we'll need an approach which also works for older versions. How about creating our own variable somewhere in which we store the instances of the extensions. 

For example, we could add a static var to `Gravity_Flow_Common` and then override `__construct()` in `Gravity_Flow_Extension` and `Gravity_Flow_Feed_Extension` to populate it e.g.

```
public function __construct() {
    parent::__construct();
    Gravity_Flow_Common::$_extensions[ get_class( $this ) ] = $this;
}
```

richardW8k (changes requested):

includes/class-extension.php
```diff
@@ -209,6 +217,31 @@ public function app_settings_tab() {
 	 */
 	public function render_uninstall() {
 
+		if ( rgget( 'page' ) == 'gravityflow_settings' && rgget( 'view' ) == 'uninstall' ) {
+			$icon        = array( 'icon' => $this->get_menu_icon() );
+			$icon_markup = GFCommon::get_icon_markup( $icon, 'dashicon-admin-generic' );
+			?>
+			<form action="" method="post" class="gform-settings-panel gform-settings-panel__addon-uninstall">
+				<?php wp_nonce_field( 'uninstall', 'gf_addon_uninstall' ); ?>
```
richardW8k, line 225: I think we should change these nonce args here so they are unique to Gravity Flow.

includes/class-feed-extension.php
```diff
@@ -211,6 +219,31 @@ public function app_settings_tab() {
 	 * Override this function to customize the markup for the uninstall section on the plugin settings page
 	 */
 	public function render_uninstall() {
+		
+		if ( rgget( 'page' ) == 'gravityflow_settings' && rgget( 'view' ) == 'uninstall' ) {
+			$icon        = array( 'icon' => $this->get_menu_icon() );
+			$icon_markup = GFCommon::get_icon_markup( $icon, 'dashicon-admin-generic' );
+			?>
+			<form action="" method="post" class="gform-settings-panel gform-settings-panel__addon-uninstall">
+				<?php wp_nonce_field( 'uninstall', 'gf_addon_uninstall' ); ?>
```
richardW8k, line 228: Same as above; we should make these unique

includes/class-extension.php
```diff
@@ -209,6 +217,31 @@ public function app_settings_tab() {
...
+				<div class="gform-settings-panel__content">
+					<div class="addon-logo dashicons"><?php echo $icon_markup; ?></div>
+					<div class="addon-uninstall-text">
+						<h4 class="gform-settings-panel__title"><?php printf( esc_html__( '%s', 'gravityforms' ), $this->get_short_title() ) ?></h4>
+						<div><?php printf( esc_html__( 'This operation deletes ALL %s settings.', 'gravityforms' ), $this->get_short_title() ) ?></div>
+					</div>
+					<div class="addon-uninstall-button">
+						<input id="addon" name="addon" type="hidden" value="<?php echo $this->get_short_title(); ?>">
```
richardW8k, line 233: How about using the slug as the input value?

includes/class-feed-extension.php
```diff
@@ -211,6 +219,31 @@ public function app_settings_tab() {
...
+				<div class="gform-settings-panel__content">
+					<div class="addon-logo dashicons"><?php echo $icon_markup; ?></div>
+					<div class="addon-uninstall-text">
+						<h4 class="gform-settings-panel__title"><?php printf( esc_html__( '%s', 'gravityforms' ), $this->get_short_title() ) ?></h4>
+						<div><?php printf( esc_html__( 'This operation deletes ALL %s settings.', 'gravityforms' ), $this->get_short_title() ) ?></div>
+					</div>
+					<div class="addon-uninstall-button">
+						<input id="addon" name="addon" type="hidden" value="<?php echo $this->get_short_title(); ?>">
```
richardW8k, line 236: If we switch to the slug in the other class we'll also need to change it here

class-gravity-flow.php
```diff
@@ -9590,5 +9598,81 @@ public function action_update_site_option_auto_update_plugins( $option, $value,
...
+		 *
+		 * @since  2.5
+		 * @return array
+		 *
+		 */
+		private static function uninstall_extensions() {
+			$installed_extensions = Gravity_Flow_Common::$_extensions;
+			$installed_extensions = array_filter( $installed_extensions, function( $extension ) {
+				return substr( $extension->get_slug(), 0, 11 ) === 'gravityflow';
+			});
```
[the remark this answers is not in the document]
richardW8k: If we have the inputs for the uninstall buttons use the add-on slugs and we update the constructors to use the slugs as the keys to the extension instances instead of the class names, we could handle the unsinstall up here instead of in the foreach e.g.

```
if ( isset( $_POST['gflow_extension_uninstall'] ) && wp_verify_nonce( $_POST['gflow_extension_uninstall'], 'gflow_extension_uninstall' ) && isset( $_POST['addon'] ) && isset( $installed_extensions[ $_POST['addon'] ] ) ) {
    $installed_extensions[ $_POST['addon'] ]->uninstall_addon();
    unset( $installed_extensions[ $_POST['addon'] ] );
    wp_redirect( $_SERVER['HTTP_REFERER'] );
}
``` 

The foreach would then become:

```
foreach ( $installed_extensions as $extension ) {
    $extension->render_uninstall();
}
```

class-gravity-flow.php
```diff
@@ -9590,5 +9598,81 @@ public function action_update_site_option_auto_update_plugins( $option, $value,
...
+
+			// Uninstall the extension and remove it from the list of installed addons on page reload.
+			if ( rgget( 'page' ) == 'gravityflow_settings' && rgget( 'view' ) == 'uninstall' ) {
```
richardW8k, line 9655: How about moving this check up to the top of the function and inverting the logic so it returns early when the `page` and `view` are wrong?

richardW8k (changes requested):

class-gravity-flow.php
```diff
@@ -9590,5 +9590,81 @@ public function action_update_site_option_auto_update_plugins( $option, $value,
...
+		 * @since  2.5
+		 * @return array
+		 *
+		 */
+		private static function uninstall_extensions() {
+			$installed_extensions = Gravity_Flow_Common::$_extensions;
+
+			if ( rgget( 'page' ) != 'gravityflow_settings' || rgget( 'view' ) != 'uninstall' ) {
+				return;
+			}
```
richardW8k, line 9645: I'd move this up above the `$installled_extensions` line

class-gravity-flow.php
```diff
@@ -9590,5 +9590,81 @@ public function action_update_site_option_auto_update_plugins( $option, $value,
...
+
+			// Uninstall the extension and remove it from the list of installed addons on page reload.
+			if ( isset( $_POST['gflow_extension_uninstall'] ) && wp_verify_nonce( $_POST['gflow_extension_uninstall'], 'gflow_extension_uninstall' ) && isset( $_POST['addon'] ) && isset( $installed_extensions[ $_POST['addon'] ] ) ) {
+				$installed_extensions[ $_POST['addon'] ]->uninstall_addon();
+				unset( $installed_extensions[ $_POST['addon'] ] );
+				wp_redirect( $_SERVER['HTTP_REFERER'] );
+			}
```
richardW8k, line 9652: If you add a break point or logging statement in here you'll find it isn’t running. It looks like the reason is the action arg of the `wp_nonce_field()` calls in the extension classes is still using `uninstall`.

class-gravity-flow.php
```diff
@@ -9590,5 +9590,81 @@ public function action_update_site_option_auto_update_plugins( $option, $value,
...
+			?>
+			<div class="gform-addons-uninstall-panel">
+			<?php
+			foreach ( $installed_extensions as $extension ) {
+				$extension->render_uninstall();
+				if ( rgpost( 'addon' ) === $extension->get_slug() ) {
+					unset( Gravity_Flow_Common::$_extensions[ $index ] );
+					$extension->uninstall_addon();
+					wp_redirect($_SERVER['HTTP_REFERER']);
+				}
```
richardW8k, line 9663: This should be removed.

includes/class-common.php
```diff
@@ -19,6 +19,8 @@
  */
 class Gravity_Flow_Common {
 
+	public static $_extensions = array();
```
richardW8k, line 22: This will want a docblock. Something like "array of extension instances with the slugs as keys"

includes/class-extension.php
```diff
@@ -49,6 +49,14 @@ abstract class Gravity_Flow_Extension extends GFAddOn {
 	 */
 	public $license_key = '';
 
+	/**
+	 * Class constructor.
+	 */		
+	public function __construct() {
+		parent::__construct();
+		Gravity_Flow_Common::$_extensions[ get_class( $this ) ] = $this;
```
richardW8k, line 57:
```suggestion
		Gravity_Flow_Common::$_extensions[ $this->get_slug() ] = $this;
```

includes/class-extension.php
```diff
@@ -209,9 +217,34 @@ public function app_settings_tab() {
 	 */
 	public function render_uninstall() {
 
+		if ( rgget( 'page' ) == 'gravityflow_settings' && rgget( 'view' ) == 'uninstall' ) {
+			$icon        = array( 'icon' => $this->get_menu_icon() );
+			$icon_markup = GFCommon::get_icon_markup( $icon, 'dashicon-admin-generic' );
+			?>
+			<form action="" method="post" class="gform-settings-panel gform-settings-panel__addon-uninstall">
+				<?php wp_nonce_field( 'uninstall', 'gflow_extension_uninstall' ); ?>
```
richardW8k, line 225:
```suggestion
				<?php wp_nonce_field( 'gflow_extension_uninstall', 'gflow_extension_uninstall' ); ?>
```

includes/class-extension.php
```diff
@@ -209,9 +217,34 @@ public function app_settings_tab() {
...
+					<div class="addon-logo dashicons"><?php echo $icon_markup; ?></div>
+					<div class="addon-uninstall-text">
+						<h4 class="gform-settings-panel__title"><?php printf( esc_html__( '%s', 'gravityforms' ), $this->get_short_title() ) ?></h4>
+						<div><?php printf( esc_html__( 'This operation deletes ALL %s settings.', 'gravityforms' ), $this->get_short_title() ) ?></div>
+					</div>
+					<div class="addon-uninstall-button">
+						<input id="addon" name="addon" type="hidden" value="<?php echo $this->get_slug(); ?>">
+						<button type="submit" aria-label="<?php printf( esc_html__( 'Uninstall %s', 'gravityforms'), $this->get_short_title() ); ?>" name="uninstall_addon" value="uninstall" class="button uninstall-addon red" onclick="return confirm('<?php echo esc_js( $this->uninstall_confirm_message() ); ?>');" onkeypress="return confirm('<?php echo esc_js( $this->uninstall_confirm_message() ); ?>');">
+							<i class="dashicons dashicons-trash"></i>
+							<?php esc_attr_e( 'Uninstall', 'gravityforms' ); ?>
```
richardW8k, line 236: the text domains for the translatable strings should be changed to `gravityflow`

includes/class-extension.php
```diff
@@ -209,9 +217,34 @@ public function app_settings_tab() {
...
+						</button>
+					</div>
+				</div>
+			</form>
+			<?php
+			return;
+		}
+
 		?>
 		<form action="" method="post">
```
richardW8k, line 246: do we still need the uninstall button on the extension settings page? If we follow the approach used by the official Gravity Forms add-ons it should be removed.

includes/class-feed-extension.php
```diff
@@ -49,6 +49,14 @@ abstract class Gravity_Flow_Feed_Extension extends GFFeedAddOn {
 	 */
 	public $license_key = '';
 
+	/**
+	 * Class constructor.
+	 */		
+	public function __construct() {
+		parent::__construct();
+		Gravity_Flow_Common::$_extensions[ get_class( $this ) ] = $this;
```
richardW8k, line 57: lets use the slug here as well

includes/class-feed-extension.php
```diff
@@ -211,10 +219,35 @@ public function app_settings_tab() {
 	 * Override this function to customize the markup for the uninstall section on the plugin settings page
 	 */
 	public function render_uninstall() {
+
```
richardW8k, line 222: this section will need the same changes as the other class

richardW8k (changes requested):

class-gravity-flow.php
saifsultanc: @richardW8k We do need it for the uninstall to take place, right? Also, need to unset/remove the extension from the `$_extensions` array when we are uninstalling the extension?

saifsultanc: resolved with this note: https://github.com/gravityflow/gravityflow/pull/419#discussion_r688920105

richardW8k (approved):

stevehenty (changes requested):
Can we make use of the uninstall_message method added in https://github.com/gravityforms/gravityforms/pull/1985 ?
Also, this PR removes the uninstall buttons for the extensions from the Forms->Settings page. Was this intentional?

PR #420, opened 2021-07-31
Ensure Notification Steps Apply Pre Submission Filter

speerface:
## Description
The GPPA crew submitted a bug report noting that when they use a Workflow Notification Step (vs. a built-in Gravity Forms notification), fields display their value instead of their label. 

After some digging, it turns out this is due to the fact that we don't pass the form object through the `gform_pre_submission_filter` before evaluating it. As such, none of the filters GPPA applies to ensure labels are displayed actually wind up being applied.

## Testing instructions
1. Add a form with a single text field an a radio field.
2. Using GPPA, set the radio field to use dynamic choices. 
3. Preview the form, and submit it with the default Notification enabled. 
4. Check the email and ensure you see the Label for the value you chose in the Radio field.
5. Now edit the form settings and disable the default notification. 
6. Add a Workflow Notification step, and choose the default admin notification as the one to send. 
7. Submit the form again. 
8. Check the email and ensure you still see the Label being displayed, and not the value.

## Checklist:
- [x] I've tested the code.


includes/steps/class-step-notification.php
```diff
@@ -106,6 +106,9 @@ function process() {
 
 		$form = $this->get_form();
 
+		// pre submission filter
+		$form = gf_apply_filters( array( 'gform_pre_submission_filter', $form['id'] ), $form );
```
richardW8k, line 110: I'm not really a fan of using the GF submission time filters as they are also often used to manipulate the `$_POST` which could impact other steps. In the past we've added context specific filters for GP to use e.g https://github.com/gravityflow/gravityflowformconnector/pull/32

speerface: Aha, ok yep that makes sense. I'll update this here as well!

speerface: Okie dokie, give this another look and let me know what you think! Tried to keep the filter name generic-ish to the notification step so it can be called whenever it's needed (vs. being named specifically for the "pre send" point of execution).

richardW8k (approved):
The version for the since tags should be 2.7.5. Unlike Gravity Forms the Gravity Flow version is bumped immediately following release, so the version you find in gravityflow.php is the one to use with the since tags, just remove `-dev`

includes/steps/class-step-notification.php
```diff
@@ -126,6 +126,28 @@ function process() {
...
+	public function filter_form( $form, $entry ) {
+		/**
+		 * Allows the form to be modified before it is processed.
+		 *
+		 * @since 2.7.6
+		 *
+		 * @param array $form  The form to be processed.
+		 * @param array $entry The entry being processed.
+		 */
+		return apply_filters( 'gravityflow_notification_form', $form, $entry );
```
Idealien, line 148: To be consistent with most of our other filters, a 3rd parameter, $this, as the $step should be included to support customizations requiring unique per step functionality. As example from the step->send_notification function:
```
$notification = apply_filters( 'gravityflow_notification', $notification, $form, $entry, $this );
```

Speaking of which, should this new gravityflow_notification_form filter also go there to support same issue resolution on approval/user input notifications? Currently there wouldn't be any way for GPPA to hydrate the form with gravityflow_notification alone.

Idealien (changes requested):

claygriffiths:
Thanks for this PR! I tested it on my end and it works as intended. 🙌

Idealien:
Follow-up testing confirmed that the scope of issue also applies for any  of the feed-based add-on step types as well.
Implementation will be handled via #422 that addresses both notifications and steps through new filters for gravityflow_{$this->get_type()}_form and gravityflow_step_form.

PR #424, opened 2021-09-12
Disable likert fields on the user input step when not selected as editable

richardW8k:
## Description
Fixes: gravityflow/backlog/issues/98

This fixes an issue where likert fields are editable on the user input step when they should be display only.

The issue is caused by `GF_Field_Likert::get_value_entry_detail()`, in the Survey add-on, returning the front-end inputs for the entry detail page. `GF_Field_Likert::get_field_input()` includes a location check which results in the `disabled` attribute being added to the inputs on the entry detail page, so the fix here is to make the field think our user input step is the entry detail page by setting its `_is_entry_detail` property to `true`.

## Testing instructions
- Create a form with a text field and two likert fields
- Create a user input step with your account as the assignee andd the second likert field selected as editable
- Preview form
- Add values to the text field and the first likert field
- Submit form
- Go to the workflow inbox page
- Click on entry
- Find the first likert field is not editable
- Find the second likert field is editable

N/A

## Documentation Changes
N/A

## Checklist:
- [x] I've tested the code.


saifsultanc (approved):
This looks great! Thanks @richardW8k

PR #433, opened 2021-11-28
Inbox Refactor to AG Grid

faction23:
## Description

Closes: https://github.com/gravityflow/backlog/issues/78

This is a mammoth PR, sorry about that. This implements the refactor of the Inbox to a system that uses [AG Grid](https://www.ag-grid.com/) enabling support for all the features outlined in the issue above. It also brings the block functionality into core, among other things.

Since we have already been testing and qa'ing this for a long time, this is more of a straight code review situation.

## Testing instructions
 
We have been testing and documenting issues for this over the duration of the work [here](https://docs.google.com/spreadsheets/d/1YxhABlF6z_H4uuEqLjWGhEK4D_wu_kJ1Mmbv6gpGzqw/edit?pli=1#gid=0)

Referring to the associated issues above and working through the features is a great way to do one more round of testing.

This pr brings in javaScript testing using Jest and applies it to the Inbox js code, but has some room for increased coverage. This will happen when we bring in the updated gulp tasks system from our npm libraries after we merge this code. 

Currently a lot of existing tests will need to be updated on this branch and we will patch those before merging.

## Documentation Changes

Many documentation updates will be needed for this release, including all new screenshots and explanations of all the new features.

## Checklist:
- [ ] I've tested the code.
- [ ] My code follows the WordPress code style. 
- [ ] My code follows the accessibility standards. 
- [ ] My code follows the inline documentation standards.


includes/inbox/models/class-task.php
```diff
@@ -0,0 +1,593 @@
...
+			return strtotime( $value );
+		}
+
+		return $value;
+	}
+
+	/**
+	 * Formats the actions for the action column.
+	 *
+	 * @param Gravity_Flow_Step $step The current step.
```
stevehenty, line 414:
```suggestion
	 * @param \Gravity_Flow_Step $step The current step.
```

includes/inbox/models/class-task.php
```diff
@@ -0,0 +1,593 @@
...
+	 *
+	 * @return string
```
stevehenty, line 416:
```suggestion
	 * @return array
```

includes/inbox/models/class-task.php
```diff
@@ -0,0 +1,593 @@
...
+		 * @since 2.5.6
+		 *
+		 * @param string $url_entry The entry URL.
+		 * @param string $entry     The current entry.
+		 * @param string $args      The inbox page arguments.
+		 * @param array  $form      The form object for the current entry.
+		 *
+		 * @return string
+		 */
+		$url_entry = apply_filters( 'gravityflow_entry_url_inbox_table', $url_entry, $entry, $args, $form );
```
stevehenty, line 566: It looks like neither of these filters is being used. Can we use them? If not, we'll need to deprecate them and perhaps find an alternative.

composer.json
```diff
@@ -34,5 +40,8 @@
     "wp-coding-standards/wpcs": "2.3.0",
     "bordoni/phpass": "dev-main",
     "yoast/phpunit-polyfills": "1.0.1"
+  },
+  "require": {
+    "league/container": "2.4.1"
```
stevehenty, line 45: Can we use the Gravity Forms class?

stevehenty (changes requested):
Great stuff! I left a few comments. The version will be 2.8 so I think we can set that now when we add the docblocks.

css/admin.css
```diff
@@ -108,7 +105,50 @@ AND ANY CHANGES MADE HERE WILL BE OVERWRITTEN.
  * Not to be used more unless absolutely necessary.
  */
 
-/* Base */
+/* -----------------------------------------------------------------------------
+ *
+ * Variables
+ *
+ * This file is just a clearing-house.
```
Idealien, line 112: Most of the section blocks have this statement - Is it default or generated by part of build process that could be removed or have single statement at top of file if it applies to all?

includes/ajax/class-endpoint.php
```diff
@@ -0,0 +1,74 @@
+<?php
+
+namespace Gravity_Flow\Gravity_Flow\Ajax;
+
+use Gravity_Flow\Gravity_Flow\Models\Model;
+use \WP_REST_Controller;
+
+abstract class Endpoint extends WP_REST_Controller {
+
+	const NAMESPACE = 'gf/v2';
```
Idealien, line 10: Should this be part of a Gravity Flow namespace?

stevehenty: In general, I like to keep Gravity Flow endpoints as part of Gravity Forms and use `workflow` as the sub-namespace.

css/admin.css
faction23: Ah yes, so css comments are left in unminified code, this comes from the fact we are commenting css partial for devs. We would have to do some conditional removal of comments as we want the first comment blocks, but then would want to strip others. Since these are machine generated we always warn people at top to not edit these.

includes/inbox/models/class-task.php
```diff
@@ -0,0 +1,593 @@
...
+		$grid     = 'inbox';
+		$type     = 'default';
+		$location = 'admin';
+
+		if ( is_admin() ) {
+			$user_id = get_current_user_id();
+			$user    = sprintf( 'user_id_%s', $user_id );
+		} else {
+			$location   = get_the_ID();
+			$filter_key = $this->api::get_inbox_filter_key( $args );
```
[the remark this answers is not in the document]
speerface: This should be resolved now; I was accidentally testing in 7.0 instead of 5.6. This and a few other errors should be resolved!

composer.json
speerface: All done!

PR #435, opened 2021-12-07
Fix HTML Field Conditional Logic on User Input Workflow Step

saifsultanc:
## Description
Re: [HS#18933](https://secure.helpscout.net/conversation/1694288950/18933?folderId=3509658)
For HTML fields the field conditional logic is not honored by the User Input step unless the step is submitted.

## Testing instructions
1. Create a form with text field A, and HTML fields B and C. Add content to HTML fields.
2. Update HTML field B with conditional logic to display only if text field A is empty.
3. Update HTML field C with conditional logic to display only if text field A is not empty.
4. Create a Workflow User Input Step with text field A as editable.
5. Create an entry for the form.
6. Go to the User Input step for the entry, confirm the conditional logic does not conform for the HTML fields B & C if we are editing content on the text field A.
7. Switch to this branch, and repeat step 6. Confirm the conditional logic does work as expected now.

N/A
 
## Screenshots
N/A

## Documentation Changes
N/A

## Checklist:
- [x] I've tested the code.


Idealien (changes requested):
The branch does complete the tests described, however if the display field setting is set to not show the html fields, the new branch does not respect that which should take priority over conditional logic. Master branch does not exhibit this issue.

Idealien (changes requested):
Display fields working well in 99% of cases. One remaining to address - When the input field which html has set conditional logic against is not in the editable (or display) list, the current branch does not consider the conditional logic, whereas master currently does.

PR #437, opened 2021-12-17
Fix Start step type display field settings not evaluating properly for entry creator

saifsultanc:
## Description
The entry creator who is not assigned workflow step(s) sees all fields of the entry, depending on where the workflow is at the moment. Any display fields restrictions applied on the Start step should be what the entry creator should be seeing all the time unless he is assigned a workflow step, or workflow completes (in which case the Complete Step setting would apply).

On code, the current check that takes place [here](https://github.com/gravityflow/gravityflow/blob/master/includes/pages/class-entry-detail.php#L864) to decide whether to use the current step settings or start step settings (for non-assignees) is to see whether the `$_POST` value is set. This will always return `true`. Even for the case of the entry creator (non-assignee), albeit an empty array. The check that we should be doing is to check for empty value, as updated on this PR.

This resolves [gravityflow/backlog#167](https://github.com/gravityflow/backlog/issues/167).

## Testing instructions
- Import this form or create a similar one which has multiple fields, display step settings limiting to a subset different from two steps (either approval/user input) that have different display/editable fields and not assigned to the entry creator.
[gravityforms-export-2021-12-08.zip](https://github.com/gravityflow/gravityflow/files/7734628/gravityforms-export-2021-12-08.zip)
- Submit the form as a contributor or user without view all permission.
- Go to a status page via block/shortcode on the front-end. Compare that fields being displayed do not match step settings for start field.
- Login as different user to complete the first step.
- Re-login as entry creator and note that the fields on display have changed - matching to display settings of step 2.
- If you repeat the above the display fields will change based on current step up until workflow complete at which point the complete step display field settings apply.
- Repeat after switching to this branch, and confirm the display fields are not appearing as expected.

N/A
 
## Screenshots
N/A

## Documentation Changes
N/A

## Checklist:
- [x] I've tested the code.


Idealien (approved):
Marking as approved - Please include change_log.txt entry before merge.

